PR #155, opened 2022-10-10
Improve adding Notebooks to the Documentation

AnesBenmerzoug:
This PR changes the way we add notebooks to the documentation.

It adds a custom extension that hooks into the `env-before-read-docs` sphinx build event and copies the notebooks before the docs' source files are read.

### DIFF
--- a/.github/workflows/tox.yaml
+++ b/.github/workflows/tox.yaml
@@ -64,9 +64,7 @@ jobs:
         key: tox-${{ github.ref }}-${{ runner.os }}-${{ hashFiles('tox.ini') }}
         path: .tox
     - name: Build Docs
-      run: |
-        cp notebooks/*.ipynb docs/examples/
-        tox -e docs
+      run: tox -e docs
     - name: Save built docs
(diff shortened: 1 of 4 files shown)


mdbenito (commented):
Silly question: What's the workflow after this PR? We would like changes in the notebook directory to be automatically reflected in the documentation. If sphinx autobuild cannot listen for changes outside of the docs folder, we could at least use a file watcher in pycharm or whatever to update files. So sphinx must pay attention to notebooks as well. However I see a --ignore argument whose purpose I don't understand...

AnesBenmerzoug:
@mdbenito The workflow after this PR is that the notebooks will be automatically copied at the of beginning of the sphinx build process.
The `--ignore` argument is used to prevent sphinx-autobuild from getting stuck in an infinite loop of rebuilding the docs because each time the notebooks get copied it detects them as new files and triggers the build. 

Currently using sphinx-autobuild is only useful for manually written rst files. Perhaps once we replace the `update_docs.py` script with a builtin sphinx extension it will also be useful for the modules' documentation.

PR #161, opened 2022-10-11
Doc/fixes

mdbenito:
Sorry for this monster PR... :/ 

## Documentation

* docstrings: formatting, wording, missing parameters, inconsistencies, docs out of sync with functionality
* main docs: sorting, moving things around, adding more explanations and some theory, add copyright notice
* notebooks: reviewed all 3 notebooks on Shapley values, added cross-references, hopefully (😅) improved the text

## Documentation build
* Hidden boilerplate cells in notebooks, but see #163 
* Make `update_docs` a bit configurable and only update newer files
* Add sphinx extension for external links (introducing new directives like `:gh:`, `:issue:` and `:tfl:` to construct standardised links to  external resources)

## Random stuff
* A bugfix and unit test that slipped in, in a726b21806dc6d3c8223e0d54451d7fbba96750d (sorry).
* Add support for math with $ and $$ in rst files
* Add ad hoc argument `stratify` to Dataset.from_sklearn() so as to simplify iris example
* Pin some dependencies, add others for dev environment
* Parallelise exact combinatorial shapley computation
* Use exact combinatorial for knn test
* Integrate knn shapley into `compute_shapley_values`


src/pydvl/utils/dataset.py
```diff
@@ -172,17 +202,28 @@ def from_sklearn(
         data: Bunch,
         train_size: float = 0.8,
         random_state: Optional[int] = None,
+        stratify: bool = False,
```
AnesBenmerzoug, line 205: This might be confusing for people expecting the same interface as sklearn [`train_test_split`](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.train_test_split.html)'s interface where the `stratify` argument is expected to be an array of the same size as the labels and not a boolean variable.

src/pydvl/shapley/__init__.py
```diff
@@ -113,7 +141,7 @@ def compute_shapley_values(
     )
 
     if val_std is None:
-        df["data_value_std"] = np.nan
+        df["data_value_std"] = np.nan  # FIXME: why NaN? stddev of a constant RV is 0
```
AnesBenmerzoug, line 144: I used NaN here by force of habit and to make sure that the values are ignored when plotting the values (which isn't the case with 0). Still I think 0 makes more sense.

notebooks/shapley_utility_learning.ipynb
```diff
@@ -0,0 +1,651 @@
...
+    "    dul_utility = DataUtilityLearning(\n",
+    "        u=utility, training_budget=budget, model=MLPRegressor(**mlp_kwargs)\n",
+    "    )\n",
+    "\n",
+    "    dul_df = compute_shapley_values(\n",
+    "        u=dul_utility,\n",
+    "        mode=\"truncated_montecarlo\",\n",
+    "        n_jobs=min(len(dataset), available_cpus()),\n",
+    "        progress=False,\n",
+    "        max_iterations=max(training_budget_values),\n",
```
AnesBenmerzoug, line 299: By setting `max_iterations=max(training_budget_values)` aren't we just reducing the effect of DUL as the training budget increases? Wouldn't be better to set it to a higher value, for e.g. `max_iterations= 2 * max(training_budget_values)` or even `max_iterations=2 ** len(u.data)`, to make sure that the only change happens because of DUL and not because truncated montecarlo shapley?

mdbenito: You are of course right. My sleepy brain interpreted max_iterations as the training budget itself! 🤦🏽 Also, it doesn't make sense to use TMCS if using a hard limit on iterations

mdbenito: OTOH, something is wrong with combinatorial MC...

AnesBenmerzoug (approved):

AnesBenmerzoug (approved):
The PR looks good to me.
Perhaps you could add changelog entries for the changes in this PR.
But other than that, feel free to merge it.

PR #162, opened 2022-10-12
Add more badges to readme

AnesBenmerzoug:
This PR adds more badges (latest version on pypi, supported python version, license) to the readme. 

Here's what the badges look like:

image


mdbenito:
yes! badges! me likey

PR #176, opened 2022-10-20
Fix chunkification of data in MapReduceJob

### LINKED ISSUE #148: MapReduceJob: chunkification of data can fail
If `len(data) > n_jobs`, then only one chunk is produced and `n_jobs-1` jobs receive empty inputs. To test, try the following in `data_utility_learning.ipynb`:

```python
import math
from numpy import ndarray
from pydvl.reporting.scores import sort_values
from pydvl.utils import MapReduceJob, maybe_progress, powerset

def test(u:Utility, n_jobs: int = 1, progress: bool=False):
    n = len(u.data)

    def job(indices:ndarray) -> ndarray:
        simplefilter("ignore", category=ConvergenceWarning)
        print(f"Worker running with indices: {indices}")
        local_values = np.zeros(n)
        for i in indices:
            subset = np.setxor1d(u.data.indices, [i], assume_unique=True)
            for s in maybe_progress(
                powerset(subset),
                progress,
                desc=f"Index {i}",
                total=2 ** (n - 1),
                position=0,
            ):
                local_values[i] += (u({i}.union(s)) - u(s)) / math.comb(n - 1, len(s))
        return local_values / n

    def reductor(results):
        return np.array(results).sum(axis=0)

    map_reduce_job = MapReduceJob(map_func=job,
                                  reduce_func=reductor,
                                  chunkify_inputs=True,
                                  n_jobs=n_jobs)
    values = map_reduce_job(u.data.indices)[0]
    return sort_values({u.data.data_names[i]: v for i, v in enumerate(values)})
```

then run it with:

```python
test(utility, max(1, available_cpus()-1))
```

Running instead the following works:

```python
test(utility, max(1, min(len(utility.data), available_cpus()-1)))
```

AnesBenmerzoug:
Closes #148

It fixes a bug in the chunkification of data.

It also fixes another bug in the `_backpressure` method and adds a regression test for it too.

### DIFF
--- a/src/pydvl/utils/parallel/map_reduce.py
+++ b/src/pydvl/utils/parallel/map_reduce.py
@@ -226,7 +226,7 @@ class MapReduceJob(Generic[T, R]):
         self, jobs: List[ObjectRef], n_dispatched: int, n_finished: int
     ) -> int:
         if (n_in_flight := n_dispatched - n_finished) > self.max_parallel_tasks:
-            wait_for_num_jobs = max(1, n_in_flight - self.max_parallel_tasks)
+            wait_for_num_jobs = n_in_flight - self.max_parallel_tasks
             self.parallel_backend.wait(
                 jobs, num_returns=wait_for_num_jobs, timeout=self.timeout
             )
--- a/src/pydvl/utils/parallel/map_reduce.py
+++ b/src/pydvl/utils/parallel/map_reduce.py
@@ -238,12 +238,23 @@ class MapReduceJob(Generic[T, R]):
     @staticmethod
     def _chunkify(data: Sequence[T], num_chunks: int) -> Iterator[Sequence[T]]:
         # Splits a list of values into chunks for each job
+        if num_chunks == 0:
+            raise ValueError("Number of chunks should be greater than 0")
+
         n = len(data)
-        chunk_size = n // num_chunks
-        remainder = n % num_chunks
+        if num_chunks > n:
+            chunk_size = 1
+            remainder = 0
+        else:
+            remainder = n % num_chunks
+            if remainder > 0:
+                num_chunks -= 1
+            chunk_size = (n - remainder) // num_chunks
         for i in range(num_chunks):
             start_index = i * chunk_size
             end_index = min(start_index + chunk_size, n)
+            if start_index >= end_index:
+                return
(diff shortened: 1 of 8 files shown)


src/pydvl/utils/parallel/map_reduce.py
```diff
@@ -226,9 +226,7 @@ def _backpressure(
         self, jobs: List[ObjectRef], n_dispatched: int, n_finished: int
     ) -> int:
         if (n_in_flight := n_dispatched - n_finished) > self.max_parallel_tasks:
-            wait_for_num_jobs = max(1, self.max_parallel_tasks // 2)
-            # num_returns cannot be bigger than length of provided list
-            wait_for_num_jobs = min(wait_for_num_jobs, n_in_flight)
+            wait_for_num_jobs = max(1, n_in_flight - self.max_parallel_tasks)
```
mdbenito, line 229: If `n_in_flight > max parallel tasks`, then the difference will always be >=1 and this `min()` is unnecessary.

But more importantly, if below the `timeout` is set, then `ray.wait()` might return before `wait_for_num_jobs` jobs are done. Shouldn't we check instead the returned lists to see the actual number of finished jobs?

src/pydvl/utils/parallel/map_reduce.py
```diff
@@ -226,9 +226,7 @@ def _backpressure(
...
             self.parallel_backend.wait(
```
mdbenito, line 230: If the `timeout` is set, then `ray.wait()` might return before `wait_for_num_jobs` jobs are done. Shouldn't we check instead the returned lists to see the actual number of finished jobs?

src/pydvl/utils/parallel/map_reduce.py
AnesBenmerzoug: Nice catch! I will fix it right away

src/pydvl/utils/parallel/map_reduce.py
AnesBenmerzoug: True, I will fix that. But then we should probably switch the condition for a while loop to prevent sending more than the maximum allowed number of jobs.

mdbenito: Closing, but see related #177

Issue #121, opened 2022-09-05
Introduce concept  of `Sampler` to determine order of subsets


AnesBenmerzoug:
In [Improving Cooperative Game Theory-Based Data Valuation Via Data Utility Learning](https://arxiv.org/pdf/2107.06336v2.pdf), the authors use a sampler $V_{samp}$ to choose for which subsets the utility will be computed and subsequently used to fit a model. As we already know, depending on which samples we use to train a given model the performance may greatly differ.   

I think that pyDVL could benefit from having different samplers overall and not jut for Data Utility Learning.

AnesBenmerzoug:
One interesting class of samplers that we could introduce are samplers based on [low-discrepancy sequences](https://en.wikipedia.org/wiki/Low-discrepancy_sequence) such as the Sobol sequence. This would of course convert the Monte Carlo methods we're using to Quasi-Monte Carlo methods and with that there may be some downsides (e.g. convergence rate and guarantees)

PR #181, opened 2022-11-03
Fix powerset sampling

mdbenito:
This pr undoes all the weird shit I did to sample from a powerset (wtf was all that about??)  and changes the test to do something more natural to me, which is to check the frequencies at which elements (roughly 2**-n) and at which subset sizes (roughly comb(n,k) appear.

I also removed the distribution parameter because it was not being used and all MC sums use a uniform distribution. I know we have #121 but I think we can add any arguments / methods / classes later if needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Simplified and fixed powerset sampling
+- Simplified and fixed powerset sampling and testing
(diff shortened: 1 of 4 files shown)


AnesBenmerzoug:
I had a quick look at the changes and this looks good to me.

It's only missing an entry in the changelog.

AnesBenmerzoug (approved):

Issue #192, opened 2022-11-14
Implement Owen sampling


mdbenito:
And its antithetic variant, as described in:

_Okhrati, Ramin, and Aldo Lipani. ‘A Multilinear Sampling Algorithm to Estimate Shapley Values’. In 2020 25th International Conference on Pattern Recognition (ICPR), 7992–99. IEEE, 2021. https://doi.org/10.1109/ICPR48806.2021.9412511._

PR #194, opened 2022-11-21
Owen sampling

mdbenito:
### Description

This monster PR (again! 🤦🏽) does a bunch of things (to be discussed in our meeting)

The branch is a misnomer.

### Changes

- Implements Owen sampling, cf #192  (but some things do not work, see #196)
- Refactoring, renaming, cleaning up
- Documentation improvements, hover popups for RTD, sphinx bibtex support
- Homogenisation of tests, some speed improvements

### Checklist

- [x] Updated Changelog


src/pydvl/value/loo/naive.py
```diff
@@ -20,8 +18,9 @@ def naive_loo(
     """
 
     values = {i: 0.0 for i in u.data.indices}
+    all_indices = set(u.data.indices)
     for i in maybe_progress(data.indices, progress):  # type: ignore
-        subset = np.setxor1d(u.data.indices, [i], assume_unique=True)
+        subset = all_indices.difference({i})
```
AnesBenmerzoug, line 23: Is this faster than setxor1d? I assumed that that was the reason why that was used.

tests/shapley/test_montecarlo.py
```diff
@@ -1,229 +1,231 @@
...
+    check_values(values, exact_values, rtol=rtol)
+    check_total_value(u, values, rtol=rtol)  # FIXME, could be more than rtol
 
 
 @pytest.mark.parametrize(
-    "a, b, num_points, fun, score_type, max_iterations, total_atol",
+    "a, b, num_points", [(2, 3, 24)]  # training set will have 0.3 * 24 ~= 7 samples
+)
+@pytest.mark.parametrize(
+    "scorer, total_atol", [("r2", 0.2), (squashed_r2, 0.1), (squashed_variance, 0.1)]
```
mdbenito, line 119: remove "r2"

src/pydvl/value/loo/naive.py
mdbenito: At any rate it is clearer, and speed is not of the essence here.

docs/30-data-valuation.rst
```diff
@@ -23,118 +23,380 @@ interest, but a function of three factors:
 
 pyDVL collects algorithms for the computation of data values in this sense,
 mostly those derived from cooperative game theory. The methods can be found in
-the modules :mod:`~pydvl.shapley` and :mod:`~pydvl.loo`, supported by
+the package :mod:`~pydvl.value`, with support from modules
 :mod:`pydvl.utils.dataset` and :mod:`~pydvl.utils.utility`, as detailed below.
 
+.. warning::
+   Be sure to read the section on
+   :ref:`the difficulties using data values <problems of data values>`.
```
AnesBenmerzoug, line 31: Minor nit: I think an `of` is missing between `difficulties` and `using`

docs/30-data-valuation.rst
```diff
@@ -23,118 +23,380 @@ interest, but a function of three factors:
...
+.. code-block:: python
+
+   from pydvl.utils import Dataset, Utility
+   from pydvl.value.shapley import compute_shapley_values
+   model = ...
+   dataset = Dataset(...)
+   utility = Utility(data, model)
+   df = compute_shapley_values(
+           u=utility, mode="owen", max_iterations=4, max_q=200
+       )
```
AnesBenmerzoug, line 264: Minor nit: The indentation is off

docs/30-data-valuation.rst
```diff
@@ -23,118 +23,380 @@ interest, but a function of three factors:
...
-   >>> df = compute_shapley_values(
+   from pydvl.utils import Dataset, Utility
+   from pydvl.value.shapley import compute_shapley_values
+
+   model = ...
+   data = Dataset(...)
+   utility = Utility(model, data)
+   df = compute_shapley_values(
            u=utility, mode="truncated_montecarlo", max_iterations=100
        )
```
AnesBenmerzoug, line 297: Minor nit: The indentation is off

docs/30-data-valuation.rst
```diff
@@ -23,118 +23,380 @@ interest, but a function of three factors:
...
+  no-free-lunch-type theorem, no single strategy can be optimal for all
+  utilities.
+
+* **Model size**: Since every evaluation of the utility entails retraining the
+  whole model on a subset of the data, large models require great amounts of
+  computation. But also, they will effortlessly interpolate small to medium
+  datasets, leading to great variance in the evaluation of performance on the
+  dedicated validation set. One mitigation for this problem is cross-validation,
+  but this would incur massive computational cost. As of v.0.3.0 there are no
+  facilities in pyDVL for cross-validating the utility (note that this would
```
AnesBenmerzoug, line 396: I think this is currently doable by subclassing the `Utility` and using cross validation in the `_utility` method.

src/pydvl/utils/parallel/map_reduce.py
```diff
@@ -54,33 +58,42 @@ def get_value(v: Union[ObjectRef, Iterable[ObjectRef], Any], *, timeout: int = 3
...
 
 class MapReduceJob(Generic[T, R]):
-    """Takes an embarrassingly parallel fun and runs it in num_jobs parallel
+    """Takes an embarrassingly parallel fun and runs it in n_jobs parallel
     jobs, splitting the data into the same number of chunks, one for each job.
 
     It repeats the process num_runs times, allocating jobs across runs. E.g.
-    if num_jobs = 90 and num_runs=10, each whole execution of fun uses 9 jobs,
-    with the data split evenly among them. If num_jobs=2 and num_runs=10, two
+    if n_jobs = 90 and num_runs=10, each whole execution of fun uses 9 jobs,
```
AnesBenmerzoug, line 65: Minor nits: `num_runs` should also be renamed to `n_runs`

src/pydvl/utils/parallel/map_reduce.py
```diff
@@ -54,33 +58,42 @@ def get_value(v: Union[ObjectRef, Iterable[ObjectRef], Any], *, timeout: int = 3
...
     :param config: Instance of :class:`~pydvl.utils.config.ParallelConfig`
         with cluster address, number of cpus, etc.
     :param n_jobs: Number of parallel jobs to run. Does not accept 0
-    :param n_runs: Number of times to run the functions on the whole data.
-    :param timeout: Amount of time in seconds to wait for remote results.
-    :param max_parallel_tasks: Maximum number of jobs to start in parallel.
+    :param n_runs: Number of times to run `map_func` and `reduce_func` on the
+        whole data.
+    :param timeout: Amount of time in seconds to wait for remote results before
+        ... TODO
```
AnesBenmerzoug, line 90: If I'm not mistaken, a TimeoutError would be raised

src/pydvl/utils/types.py
```diff
@@ -71,3 +76,72 @@ def items(self):
...
+    :param fun: The function to wrap
+    :param new_arg: The name of the argument that the new function will accept
+        (and ignore).
+    :return: A new function accepting one more keyword argument.
+    """
+    params = inspect.signature(fun).parameters
+    if new_arg in params.keys():
+        return fun
+
+    def wrapper(*args, **kwargs):
```
AnesBenmerzoug, line 98: Maybe decorating this with `functools.wraps` would be a good idea, even though it's only used internally.

src/pydvl/value/shapley/__init__.py
```diff
@@ -32,6 +38,8 @@ class ShapleyMode(str, Enum):
     CombinatorialMontecarlo = "combinatorial_montecarlo"
     PermutationMontecarlo = "permutation_montecarlo"
     TruncatedMontecarlo = "truncated_montecarlo"
+    Owen = "owen"
+    OwenAntithetic = "owen_anti"
```
AnesBenmerzoug, line 42: Minor nit: For the sake of consistency, I think `owen_anti` should be changed to `owen_antithetic`

src/pydvl/value/shapley/knn.py
```diff
@@ -19,23 +19,32 @@ def knn_shapley(
...
-    :param progress: whether to display a progress bar.
+    :param progress: Whether to display a progress bar.
 
     :return: An
         `OrderedDict <https://docs.python.org/3/library/collections.html#collections.OrderedDict>`_
         of data indices and their values.
+
+    .. rubric:: References
+
+    [1]: Jia, Ruoxi, David Dao, Boxin Wang, Frances Ann Hubis, Nezihe Merve
```
AnesBenmerzoug, line 40: I think this should use the citation style.

docs/30-data-valuation.rst
mdbenito: black seems to think it's ok. Otherwise the tests wouldn't pass, right?

docs/30-data-valuation.rst
mdbenito: ditto

docs/30-data-valuation.rst
mdbenito: Isn't that precisely proof that we don't provide a ready solution? That would be a `CrossValidatedUtility` ready to be used.

docs/30-data-valuation.rst
mdbenito: Why don't you add suggestions? like so:

```suggestion
   :ref:`the difficulties of using data values <problems of data values>`.
```

src/pydvl/utils/parallel/map_reduce.py
mdbenito:
```suggestion
    if n_jobs = 90 and n_runs=10, each whole execution of fun uses 9 jobs,
```

src/pydvl/utils/types.py
mdbenito: Thought about it, but didn't want to think about that for no particular benefit. Why do you think we should add it?

src/pydvl/value/shapley/__init__.py
mdbenito: yikes, RSI says hello... But yes, I think you are right

src/pydvl/value/shapley/knn.py
mdbenito: dammit!

src/pydvl/value/shapley/montecarlo.py
```diff
@@ -0,0 +1,521 @@
...
+    """
+    values = np.zeros(len(u.data))
+
+    q_stop = {OwenAlgorithm.Standard: 1.0, OwenAlgorithm.Antithetic: 0.5}
+    q_steps = np.linspace(start=0, stop=q_stop[method], num=max_q)
+
+    index_set = set(indices)
+    for i in maybe_progress(indices, progress, position=job_id):
+        e = np.zeros(max_q)
+        subset = np.array(list(index_set.difference({i})))
```
AnesBenmerzoug, line 410: I actually went and timed this against using numpy's setxor1d and it turns there isn't much difference.
setxor1d scales better but you need a lot of indices for it to even matter.

src/pydvl/utils/types.py
mdbenito: Forget it, it's easier to just add it 😅

docs/30-data-valuation.rst
AnesBenmerzoug: I did not know or think of using it. Thanks

```suggestion
   :ref:`the difficulties using of data values <problems of data values>`.
```

src/pydvl/utils/parallel/map_reduce.py
mdbenito: docs tell me that would be a `ray.exceptions.GetTimeoutError` but also a RayTaskError could be raised here or elsewhere if something fails. So there are many more places to warn about exceptions. Maybe we can delay documenting all exceptions raised to some other time? Also, we would have to abstract the backend somehow.

docs/30-data-valuation.rst
AnesBenmerzoug: I don't think black checks rst files. We could consider using [blacken-docs](https://pypi.org/project/blacken-docs/) to do that.

docs/30-data-valuation.rst
AnesBenmerzoug: You're right. The way you phrased it in the docs is truer.
Still, I think that gives off the idea that it is impossible without us implementing it first.
Perhaps we could showcase this possibility in a new notebook in which we could show more of these data valuation issues.

src/pydvl/utils/parallel/map_reduce.py
```diff
@@ -54,33 +58,42 @@ def get_value(v: Union[ObjectRef, Iterable[ObjectRef], Any], *, timeout: int = 3
...
 class MapReduceJob(Generic[T, R]):
-    """Takes an embarrassingly parallel fun and runs it in num_jobs parallel
+    """Takes an embarrassingly parallel fun and runs it in n_jobs parallel
     jobs, splitting the data into the same number of chunks, one for each job.
 
     It repeats the process num_runs times, allocating jobs across runs. E.g.
-    if num_jobs = 90 and num_runs=10, each whole execution of fun uses 9 jobs,
-    with the data split evenly among them. If num_jobs=2 and num_runs=10, two
+    if n_jobs = 90 and num_runs=10, each whole execution of fun uses 9 jobs,
+    with the data split evenly among them. If n_jobs=2 and num_runs=10, two
```
AnesBenmerzoug, line 66:
```suggestion
    with the data split evenly among them. If n_jobs=2 and n_runs=10, two
```

AnesBenmerzoug (approved):
Except for the few minor comments I left, this looks good to me.
Feel free to merge it.

docs/30-data-valuation.rst
mdbenito: I really need to learn to pause before writing the first that comes to my mind. Will fix

docs/30-data-valuation.rst
mdbenito: Sounds like a good idea!

src/pydvl/utils/types.py
mdbenito: It turns out it isn't. Something breaks if I do (pickling of parallel jobs, don't ask, I'm just going to ignore this 😇 )

docs/30-data-valuation.rst
mdbenito: https://github.com/appliedAI-Initiative/pyDVL/issues/198

src/pydvl/value/shapley/montecarlo.py
mdbenito: If "a lot" is millions, and the difference is in ms, then we can totally ignore this

Issue #130, opened 2022-09-20
Investigate flakiness of influence values


Xuzzo:
I have tried in several ways to show that the influence values of corrupted data-points (points for which I have flipped the label) is negative, but the results are very noisy for NNs. Investigate this issue and create a notebook that goes through such limitations of IF, as described in [this paper](https://arxiv.org/pdf/2006.14651.pdf)

PR #195, opened 2022-11-23
Cnn influence example

### LINKED ISSUE #128: Reduce memory usage in influence calculation
To calculate influence values we copy the dataset several times. This expands a lot the memory usage. Optimise the workflow to avoid it.

### LINKED ISSUE #209: IF calculation breaks with multi-dimensional input
The current implementation of IF works only with one dimensional input to the NN.

### LINKED ISSUE #221: Add notebooks requirements file
Some notebooks need specific dependencies (e.g. huggingface datasets for the imagenet notebook). In our pipeline, this is managed by tox environments, but users that do not use tox would have import errors if they only install the normal reqs. We should add a notebooks-requirements file.

Xuzzo:
### Description

This PR introduces an example where influence functions are used for CNNs. 

### Changes
- Conjugate gradient calculated implicitly through scipy LinearOperator. Closes #128 
- IF computation now supports multi-dimensional input. Closes #209
- New notebook with examples and theory
- Progress on #130 
- Closes #221 

### Checklist

- [x] Updated Changelog


src/pydvl/utils/dataset.py
```diff
@@ -465,6 +467,60 @@ def load_wine_dataset(
     )
 
 
+def load_preprocess_imagenet(
```
mdbenito, line 470: Since this is support for the notebook, I think it can be moved to the support file that you can find under `/notebooks` in develop.

requirements.txt
```diff
@@ -7,3 +7,4 @@ pymemcache
 cloudpickle
 tqdm
 matplotlib
+datasets
```
AnesBenmerzoug, line 10: Since this package is only used in the notebooks, I think it makes more sense to just add it to the `notebooks` tox environment instead.

src/pydvl/influence/general.py
```diff
@@ -189,7 +190,7 @@ def compute_influences(
     n_params = differentiable_model.num_params()
     dict_fact_algos: Dict[Optional[str], MatrixVectorProductInversionAlgorithm] = {
         "direct": lambda hvp, x: np.linalg.solve(hvp(np.eye(n_params)), x.T).T,  # type: ignore
-        "cg": lambda hvp, x: conjugate_gradient(hvp(np.eye(n_params)), x),  # type: ignore
+        "cg": lambda hvp, x: conjugate_gradient(LinearOperator((n_params, n_params), matvec=hvp), x),  # type: ignore
```
AnesBenmerzoug, line 193: What is the benefit of using `LinearOperator`?

src/pydvl/influence/model_wrappers/torch_wrappers.py
```diff
@@ -106,8 +108,13 @@ def fit(
                 if scheduler:
                     scheduler.step()
             pred_val = self.forward(x_val)
-            val_loss.append(loss(torch.squeeze(pred_val), torch.squeeze(y_val)).item())
-            train_loss.append(np.mean(batch_loss))
+            epoch_val_loss = loss(torch.squeeze(pred_val), torch.squeeze(y_val)).item()
+            mean_epoch_train_loss = np.mean(batch_loss)
```
AnesBenmerzoug, line 112: After a certain a point, it might make sense to just rely on pytorch-lightning for these things.

notebooks/notebook_support.py
```diff
@@ -258,3 +271,323 @@ def _handle_legend(scatter):
...
+    split_ds = split_ds["train"].train_test_split(
+        train_size=train_size,
+        seed=random_state,
+    )
+    train_ds = _process_dataset(split_ds["train"])
+    val_ds = _process_dataset(split_ds["test"])
+    return train_ds, val_ds, test_ds
+
+
+def save_model(model, train_loss, val_loss, model_name):
```
AnesBenmerzoug, line 366: This function is missing type hints.

notebooks/notebook_support.py
```diff
@@ -258,3 +271,323 @@ def _handle_legend(scatter):
...
+    :param model_name: model name, used for saving the files
+    """
+    torch.save(model.state_dict(), imgnet_model_data_path / f"{model_name}_weights.pth")
+    with open(
+        imgnet_model_data_path / f"{model_name}_train_val_loss.pkl", "wb"
+    ) as file:
+        pkl.dump([train_loss, val_loss], file)
+
+
+def load_model(model, model_name):
```
AnesBenmerzoug, line 381: This function is missing type hints.

notebooks/notebook_support.py
```diff
@@ -258,3 +271,323 @@ def _handle_legend(scatter):
...
+
+    :param file_name: string, file name where the object is saved
+    :return: saved object
+    """
+    with open(imgnet_model_data_path / f"{file_name}", "rb") as file:
+        results = pkl.load(file)
+    return results
+
+
+def plot_sample_images(
```
AnesBenmerzoug, line 420: This function is missing type hints.

notebooks/notebook_support.py
```diff
@@ -258,3 +271,323 @@ def _handle_legend(scatter):
...
+        ):
+            axes[img_idx, class_idx].imshow(img_data["images"])
+            axes[img_idx, class_idx].axis("off")
+            axes[img_idx, class_idx].set_title(f"img label: {class_label}")
+            if img_idx + 1 >= n_images_per_class:
+                break
+    plt.show()
+
+
+def plot_top_bottom_if_images(
```
AnesBenmerzoug, line 445: This function is missing type hints.

notebooks/notebook_support.py
```diff
@@ -258,3 +271,323 @@ def _handle_legend(scatter):
...
+
+    for plt_idx, img_idx in enumerate(top_if_idxs):
+        axes[plt_idx, 1].set_title(f"img influence: {subset_influences[img_idx]:0f}")
+        axes[plt_idx, 1].imshow(subset_images[img_idx])
+        axes[plt_idx, 1].axis("off")
+
+    plt.show()
+
+
+def plot_train_val_loss(train_loss, val_loss):
```
AnesBenmerzoug, line 476: This function is missing type hints.

notebooks/notebook_support.py
```diff
@@ -258,3 +271,323 @@ def _handle_legend(scatter):
...
+    _, ax = plt.subplots()
+    ax.plot(train_loss, label="Train")
+    ax.plot(val_loss, label="Val")
+    ax.set_ylabel("Loss")
+    ax.set_xlabel("Train epoch")
+    ax.legend()
+    plt.show()
+
+
+def get_corrupted_imagenet(dataset, fraction_to_corrupt, avg_influences):
```
AnesBenmerzoug, line 491: This function is missing type hints.

notebooks/notebook_support.py
```diff
@@ -258,3 +271,323 @@ def _handle_legend(scatter):
...
+        ).index.tolist()
+        wrong_labels = [l for l in labels if l != label]
+        for img_idx in indices_to_corrupt:
+            sample_label = np.random.choice(wrong_labels)
+            corrupted_dataset.at[img_idx, "labels"] = sample_label
+            corrupted_indices[sample_label].append(img_idx)
+    return corrupted_dataset, corrupted_indices
+
+
+def plot_influence_distribution_by_label(influences, dataset):
```
AnesBenmerzoug, line 526: This function is missing type hints.

notebooks/notebook_support.py
```diff
@@ -258,3 +271,323 @@ def _handle_legend(scatter):
...
+    for label in labels:
+        ax.hist(influences[dataset["labels"] == label], label=label, alpha=0.7)
+    ax.set_xlabel("influence values")
+    ax.set_ylabel("number of points")
+    ax.set_title("Influence distribution")
+    ax.legend()
+    plt.show()
+
+
+def plot_corrupted_influences_distribution(
```
AnesBenmerzoug, line 544: This function is missing type hints.

AnesBenmerzoug:
@Xuzzo Thanks for your work. There are quite a few changes so I think it makes sense to go through all of them during a call. 

But first, could you please update the changelog?

src/pydvl/influence/general.py
Xuzzo: It calculates matrix vector product only implicitly. This means that instead of having to store the whole matrix, it uses less memory. More info can be found on the [official scipy documentation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.LinearOperator.html)

src/pydvl/influence/model_wrappers/torch_wrappers.py
Xuzzo: Wrote [issue about this.](https://github.com/appliedAI-Initiative/pyDVL/issues/222).

Xuzzo:
Typing should be done now. I have also updated the changelog.

Xuzzo:
- Removed caching of results. Kept the one for the model
- I have added the difference in score between corrupted and simple.

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -79,16 +79,27 @@ def split_grad(
...
+            param for param in self.model.parameters() if param.requires_grad == True
+        ]
+
         grads = [
             flatten_gradient(
                 autograd.grad(
-                    self.loss(torch.squeeze(self.model(x[i])), torch.squeeze(y[i])),
-                    self.model.parameters(),
+                    self.loss(
+                        torch.squeeze(self.model(x[i].unsqueeze(0))),
```
AnesBenmerzoug, line 90: Perhaps it would make sense to check the number of dimensions at the beginning of this function and unsqueeze just like you have done in the `mvp` function.

src/pydvl/influence/general.py
```diff
@@ -68,15 +70,14 @@ def calculate_influence_factors(
...
-    :returns: A np.ndarray of size (N, D) containing the influence factors for each dimension (D) and test sample (N).
     :param progress: If True, display progress bars.
+    :returns: A np.ndarray of size (N, D) containing the influence factors for each dimension (D) and test sample (N).
     """
     if not _TORCH_INSTALLED:
         raise RuntimeWarning("This function requires PyTorch.")
     grad_xy, _ = model.grad(x, y)
-    hvp = lambda v: model.mvp(grad_xy, v, progress=progress) + lam * v
-    test_grads = model.split_grad(x_test, y_test, progress=progress)
+    hvp = lambda v: model.mvp(grad_xy, v) + lam * v
```
AnesBenmerzoug, line 79: You don't pass the progress parameter to the `mvp` method anymore?

AnesBenmerzoug:
@Xuzzo Thanks for your work! It looks better now.

The main thing I noticed is that in the imagenet notebook when you load the dataset it prints information that is useless to the notebook reader:

image

Perhaps we should call [`datasets.utils.logging.set_verbosity_error()`](https://huggingface.co/docs/datasets/package_reference/logging_methods#datasets.utils.logging.set_verbosity_error) in the `load_preprocess_imagenet` function.

src/pydvl/influence/general.py
Xuzzo: No because in conjugate gradient hvp is called several times with only one element (and so progress bar gets spammed).

notebooks/notebook_support.py
```diff
@@ -258,3 +283,314 @@ def _handle_legend(scatter):
...
+                processed_ds["normalized_images"].append(preprocess_rgb(item["image"]))
+                processed_ds["images"].append(item["image"])
+                processed_ds["labels"].append(item["label"])
+        return pd.DataFrame.from_dict(processed_ds)
+
+    if is_CI:
+        tiny_imagenet = load_dataset("Maysee/tiny-imagenet", split="valid")
+        tiny_imagenet_train = tiny_imagenet.shard(10, 0)
+        tiny_imagenet_val = tiny_imagenet.shard(10, 1)
+        tiny_imagenet_test = tiny_imagenet.shard(2, 1)
```
mdbenito, line 340: I'm not sure how HF datasets work, but even if shard() is deterministic it is not contiguous by default, so, shouldn't one first shard in train/test, and then shard the training set into train/val? 

```suggestion
        split = tiny_imagenet.shard(2, 0)
        tiny_imagenet_test = tiny_imagenet.shard(2, 1)
        tiny_imagenet_train = split.shard(5, 0)
        tiny_imagenet_val = split.shard(5, 1)
```

This should use 10% for training, 10% for validation and 50% for testing

notebooks/notebook_support.py
```diff
@@ -258,3 +283,314 @@ def _handle_legend(scatter):
...
+        if keep_labels is not None:
```
mdbenito, line 341: Maybe do this before the split to avoid doing it three times? Or is that too slow?

notebooks/notebook_support.py
```diff
@@ -258,3 +283,314 @@ def _handle_legend(scatter):
...
+    train_ds = _process_dataset(split_ds["train"])
+    val_ds = _process_dataset(split_ds["test"])
+    return train_ds, val_ds, test_ds
+
+
+def save_model(
+    model: TorchModel,
+    train_loss: List[float],
+    val_loss: List[float],
+    model_name: Literal,
```
mdbenito, line 382: Maybe you forgot the choices for the Literal type? Or did you mean `str`?

notebooks/notebook_support.py
```diff
@@ -258,3 +283,314 @@ def _handle_legend(scatter):
...
+    """
+    torch.save(model.state_dict(), imgnet_model_data_path / f"{model_name}_weights.pth")
+    with open(
+        imgnet_model_data_path / f"{model_name}_train_val_loss.pkl", "wb"
+    ) as file:
+        pkl.dump([train_loss, val_loss], file)
+
+
+def load_model(
+    model: TorchModel, model_name: Literal
```
mdbenito, line 399: Same as above about Literal and str

notebooks/notebook_support.py
```diff
@@ -258,3 +283,314 @@ def _handle_legend(scatter):
...
+    with open(
+        imgnet_model_data_path / f"{model_name}_train_val_loss.pkl", "rb"
+    ) as file:
+        train_loss, val_loss = pkl.load(file)
+    return train_loss, val_loss
+
+
+def plot_sample_images(
+    dataset: pd.DataFrame,
+    n_images_per_class=3,
```
mdbenito, line 420:
```suggestion
    n_images_per_class: int = 3,
```

notebooks/notebook_support.py
```diff
@@ -258,3 +283,314 @@ def _handle_legend(scatter):
...
+):
+    """Given the preprocessed imagenet dataset (or a subset of it), it plots \
+    a number n_images_per_class of images for each class.
+
+    :param dataset: imagenet dataset
+    :param n_images_per_class: int, number of images per class to plot
```
mdbenito, line 426:
```suggestion
    :param n_images_per_class: number of images per class to plot
```

notebooks/notebook_support.py
```diff
@@ -258,3 +283,314 @@ def _handle_legend(scatter):
...
+    for class_idx, class_label in enumerate(labels):
+        for img_idx, (_, img_data) in enumerate(
+            dataset[dataset["labels"] == class_label].iterrows()
+        ):
+            axes[img_idx, class_idx].imshow(img_data["images"])
+            axes[img_idx, class_idx].axis("off")
+            axes[img_idx, class_idx].set_title(f"img label: {class_label}")
+            if img_idx + 1 >= n_images_per_class:
+                break
+    plt.show()
```
mdbenito, line 440: I think we should change all plotting functions to accept axes and not plot, see #234 (for another time)

notebooks/notebook_support.py
```diff
@@ -258,3 +283,314 @@ def _handle_legend(scatter):
...
+    _, ax = plt.subplots()
+    ax.plot(train_loss, label="Train")
+    ax.plot(val_loss, label="Val")
+    ax.set_ylabel("Loss")
+    ax.set_xlabel("Train epoch")
+    ax.legend()
+    plt.show()
+
+
+def get_corrupted_imagenet(
```
mdbenito, line 489: This is not a "getter", but a transform. I'd call it `corrupt_imagenet` instead.

notebooks/notebook_support.py
```diff
@@ -258,3 +283,314 @@ def _handle_legend(scatter):
...
+    dataset: pd.DataFrame, fraction_to_corrupt: float, avg_influences: "NDArray"
```
mdbenito, line 490:
```suggestion
    dataset: pd.DataFrame, fraction_to_corrupt: float, avg_influences: "NDArray[np.float_]"
```

mdbenito (commented):
I've looked a bit at the supporting code but couldn't check the notebooks yet. Will do asap

Xuzzo:
MR comments should be addressed.

AnesBenmerzoug (approved):
I had another quick look at it. It looks good to me and can be merged!

mdbenito:
Hi hi,

the supporting code is better now, thanks!

I haven't had time to check it in detail yet, but the notebook:

* contains several typos, please run it through a spell checker. Some of them are recurring and I have fixed many times in the past, like "data-points". 🙏🏽 
* does not hide cells as instructed in the PR template and CONTRIBUTING.md (why?)
* Points to a whole book as reference for the main computation. At least one should mention the section or theorem.
* has a (useful) theory section which ignores the theory section in the review document draft, and is based on a paper which diverges from the usual theory of influence functions.

I understand that it makes sense to be consistent with the paper that is implemented, but a major selling point of the review and of a library like pydvl is that we provide a common framework and notation to understand any paper on influence functions. In particular, it is important to unify notation and be general enough. Koh2017 is a rather sloppy source because it fails to properly define the object that is being approximated and also diverges from the usual definitions. Custom definitions are ok, but we should strive to make the link to what is standard in the literature, I believe.

Now, the stub I wrote over a year ago is itself worse than sloppy, I'm not claiming it's any better (actually I just read it and it's pretty confusing, to put it mildly 😅). But it'd be nice to try to link both. It is ok to remain superficial in the notebook, if it is consistent with the review. Maybe you can start by mentioning the latter, and in the future work to improve that document (we should coordinate, though).

notebooks/notebook_support.py
Xuzzo: I have left it with the plot for now. I guess we can unify the interface altogether at some point.

Xuzzo:
To do, from discussion in #235:
- [ ] add notebook requirements
- [ ] check model saving
- [ ] check dataset ram usage
- [ ] Investigate and (maybe) refactor InternalDataset

PR #199, opened 2022-11-24
Allow manual triggering of pypi releases

mdbenito:
The pipeline keeps failing. It seems to be some misconfiguration, but some time it will be a transient error, so I don't think it's bad to have a manual trigger.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -30,7 +30,7 @@ jobs:
         id: get_branch_name
         if: github.ref_type == 'tag'
         run: |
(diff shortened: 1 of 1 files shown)


AnesBenmerzoug (changes requested):
This is definitely a good change. But I think what failed the pipeline is the condition on line 27. I added it and I now realize it is wrong because when a tag triggers the workflow github.ref has the tag as value.

mdbenito:
> This is definitely a good change. But I think what failed the pipeline is the condition on line 27. I added it and I now realize it is wrong because when a tag triggers the workflow github.ref has the tag as value.

But we need some check for the branch, though. We don't want a release to happen from anything other than master, not even manually.

mdbenito (commented):
lgtm at first sight

PR #214, opened 2022-12-03
ValuationResult

### LINKED ISSUE #178: Define and use return types for all Shapley methods
Instead of returning ordereddicts or tuples, use type aliases, namedtuples and dataclasses.

mdbenito:
This PR closes #178:

* Introduces a class `ValuationResult` to gather and inspect results from all valuation algorithms
* `ValuationResult` is a Sequence (supports indexing, with lists too) which can be exported to DataFrame
* Consistently uses it across all methods
* Adds tests and documentation

Things that could be done on top:

* [ ] Include an additional array `n_updates` for the number of times that an index has been updated by a Monte Carlo method. This is necessary in order to aggregate results (to have the weights needed to combine means and variances computed from sequences of different lengths)
* [ ] Add utilities to merge results. Useful e.g. for cross-validating value computations.


src/pydvl/value/results.py
```diff
@@ -0,0 +1,232 @@
...
+    cast,
+    overload,
+)
+
+import numpy as np
+
+from pydvl.utils import Dataset, SortOrder
+
+try:
+    import pandas  # Try to import here for the benefit of mypy
```
AnesBenmerzoug, line 32: Pandas is part of the requirements. I don't see how this import could fail.

src/pydvl/value/results.py
```diff
@@ -0,0 +1,232 @@
...
+    _sort_order: Optional[SortOrder] = None
+
+    def __init__(
+        self,
+        algorithm: Callable,  # BaseValuator,
+        status: ValuationStatus,  # Valuation.Status,
+        values: "NDArray[np.float_]",
+        stderr: Optional["NDArray[np.float_]"] = None,
+        data_names: Optional[Sequence[str]] = None,
+        sort: Optional[SortOrder] = None,
```
AnesBenmerzoug, line 101: Wouldn't it be simpler to just use a boolean flag since there are only 2 possible options?

src/pydvl/value/results.py
```diff
@@ -0,0 +1,232 @@
...
+        return self._sort_order is not None
+
+    @property
+    def values(self) -> "NDArray[np.float_]":
+        """The raw values, unsorted. Position `i` in the array represents index
+        `i` of the data."""
+        return self._values
+
+    @property
+    def indices(self) -> "NDArray":
```
AnesBenmerzoug, line 151:
```suggestion
    def indices(self) -> "NDArray[np.int_]":
```

src/pydvl/value/results.py
```diff
@@ -0,0 +1,232 @@
...
+    _data: Dataset
+    _names: "Union[NDArray[np.int_], NDArray[np.str_]]"
+    _stderr: "NDArray[np.float_]"
+    _algorithm: str  # TODO: BaseValuator
+    _status: ValuationStatus  # TODO: Maybe? BaseValuator.Status
+    _sort_order: Optional[SortOrder] = None
+
+    def __init__(
+        self,
+        algorithm: Callable,  # BaseValuator,
```
AnesBenmerzoug, line 96: What is the reason for passing the algorithm function instead of the algorithm name?

src/pydvl/value/loo/naive.py
```diff
@@ -1,26 +1,32 @@
-from typing import OrderedDict
-
 import numpy as np
 
-from pydvl.reporting.scores import sort_values
-from pydvl.utils import Utility, maybe_progress
+from pydvl.utils import SortOrder, Utility, maybe_progress
+from pydvl.value import ValuationResult, ValuationStatus
```
AnesBenmerzoug, line 4: I think it would be safer here to import directly from `pydvl.value.results` to avoid any possible future circular import errors.

tests/test_results.py
```diff
@@ -0,0 +1,108 @@
+import numpy as np
+import pytest
+
+from pydvl.utils.types import SortOrder
+from pydvl.value import ValuationResult, ValuationStatus
+
+
+def dummy_valuator():
+    pass
+
```
AnesBenmerzoug, line 10: I think adding a test that makes sure that the objects can be serialized and deserialized without errors would be good idea.

AnesBenmerzoug (changes requested):
Overall this looks good to me. It gives it a bit more consistency to the valuation results.
Other than the few comments I left, I don't see any issues with this change.
The only complaint I might have with this is that  it is once again a big change and I think I will create a new branch for the least core changes to avoid dealing with too many conflicts/

src/pydvl/value/results.py
mdbenito: It's a leftover. It was intended as a placeholder because I was toying with the idea of using objects for the algorithms and having a `BaseValuator`. This would then be passed here and used to query statistics or whatever. But I'm not so sure about that anymore (sklearn interfaces don't fully make sense, only localizing the cache is a good argument), and the callable with no signature is silly

src/pydvl/value/results.py
mdbenito: Good catch. But I don't think it should be a requirement. As a matter of fact, I see it's only really used in Dataset, and that could all be put into a factory `Dataset.from_dataframe()`, making pandas optional. So I'll leave this  here and create #218

src/pydvl/value/loo/naive.py
mdbenito:
```suggestion
from pydvl.value.results import ValuationResult, ValuationStatus
```

src/pydvl/value/shapley/__init__.py
```diff
@@ -5,15 +5,11 @@
...
-import warnings
 from enum import Enum
 from typing import Optional, cast
 
-import numpy as np
-import pandas as pd
-import sklearn.neighbors as skn
-
 from pydvl.utils import Utility
+from pydvl.value import ValuationResult
```
mdbenito, line 12:
```suggestion
from pydvl.value.results import ValuationResult
```

src/pydvl/value/shapley/actor.py
```diff
@@ -16,6 +16,7 @@
 from pydvl.utils.config import ParallelConfig
 from pydvl.utils.parallel.actor import Coordinator, RayActorWrapper, Worker
 from pydvl.utils.parallel.backend import init_parallel_backend
+from pydvl.value import ValuationStatus
```
mdbenito, line 19:
```suggestion
from pydvl.value.results import ValuationStatus
```

src/pydvl/value/shapley/knn.py
```diff
@@ -1,69 +1,69 @@
...
-from typing import Dict, OrderedDict, Union
+from typing import Dict, Union
 
+import numpy as np
 from sklearn.neighbors import KNeighborsClassifier, NearestNeighbors
 
-from pydvl.reporting.scores import sort_values
-from pydvl.utils import Dataset, maybe_progress
+from pydvl.utils import SortOrder, Utility, maybe_progress
+from pydvl.value import ValuationResult, ValuationStatus
```
mdbenito, line 14:
```suggestion
from pydvl.value.results import ValuationResult, ValuationStatus
```

src/pydvl/value/shapley/naive.py
```diff
@@ -1,26 +1,24 @@
...
 from pydvl.utils import (
     MapReduceJob,
     ParallelConfig,
+    SortOrder,
     Utility,
     init_parallel_backend,
     maybe_progress,
     powerset,
 )
+from pydvl.value import ValuationResult, ValuationStatus
```
mdbenito, line 16:
```suggestion
from pydvl.value.results import ValuationResult, ValuationStatus
```

tests/test_results.py
mdbenito: A very good one

src/pydvl/value/results.py
mdbenito: Yes, but a boolean would feel rather weird for "ascending" vs "descending". It is a total PITA having to import the class and to type so much, I agree. I wanted to use the string enum values, but there is no way to automatically construct a literal from them that mypy accepts, and in the end that kind of defeats the purpose since user code won't necessarily use a type checker.

mdbenito: Also, I think originally I thought that one may want to add different sorting methods, e.g. taking into account the stderr. But that's admittedly rather unlikely to be very useful.

mdbenito: I guess I'll just use a literal and do some arg checking inside sort()

AnesBenmerzoug: Before this change, the results were sorted in an ascending order by default. What I'm trying to understand is what would be the use case for sorting in a different order or for not sorting at all?

mdbenito: Being able to pick the order is just a convenience. Sort descending (resp. ascending) to iterate first over high (resp. low) value points. No sorting means you iterate over points using their indices. I have made this optional just to be more explicit about what's going on.

src/pydvl/value/results.py
mdbenito: Fixed in https://github.com/appliedAI-Initiative/pyDVL/pull/214/commits/2e3a329a35dc051645e10f480b640a94976a3a53

tests/test_results.py
mdbenito: Implemented in https://github.com/appliedAI-Initiative/pyDVL/pull/214/commits/05ac4d1cc48fe6006c975e110ed7966bbb45b03d for pickle and cloudpickle.

src/pydvl/value/results.py
mdbenito: I've experimented with Literals, but I don't like how it looks. It's a purely aesthetic thing and hence subject to debate, but since it's a rather minor issue and I need to wrap this up before next week's meeting madness, I'm going to resolve the conversation. But feel free to reference this thread in a new issue if you find the Enum too awkward.

mdbenito:
> Overall this looks good to me. It gives it a bit more consistency to the valuation results. Other than the few comments I left, I don't see any issues with this change. The only complaint I might have with this is that it is once again a big change and I think I will create a new branch for the least core changes to avoid dealing with too many conflicts/

Sorry about that! In my defence, I'll say that this time I did limit myself to a single issue... 😅

src/pydvl/value/results.py
AnesBenmerzoug: I still have to say YAGNI to this change. `ValuationResult` would be a simpler as just a container and not a container with extra functionality that can also be obtained by converting it to a pandas dataframe.

src/pydvl/value/results.py
```diff
@@ -0,0 +1,272 @@
...
+        """Sorts the indices in place by ascending value.
+
+        Repeated calls with the same sort order are no-ops. Once sorted,
+        iteration over the results will follow the order.
+
+        :param reverse: Whether to sort in descending order by value.
+        """
+
+        # Try to save time if we are already sorted in some way
+        if self._sort_order is not None:
```
AnesBenmerzoug, line 158: I think it would be better if we make this method completely stateless. It just adds complexity to the code without much benefit.

mdbenito: Speed! I can call sort multiple times without caring.

src/pydvl/value/results.py
```diff
@@ -0,0 +1,272 @@
...
+            self._indices = np.arange(0, len(self._values), dtype=np.int_)
+
+        if data_names is None:
+            self._names = np.arange(0, len(values), dtype=np.int_)
+        else:
+            self._names = np.array(data_names)
+        if len(self._names) != len(self._values):
+            raise ValueError("Data names and data values have different lengths")
+
+    def sort(self, reverse: bool = False) -> None:
```
AnesBenmerzoug, line 148: It would be nice to offer this interface instead to allow using the std in the comparison, for example
```suggestion
    def sort(self, *, key=None, reverse=False) -> None:
```
But this would mean that we cannot use argsort, which would make this slower.

src/pydvl/value/results.py
AnesBenmerzoug: Did you measure the difference? I think the stability of the sorting algorithm is more important than speed and in this case both the python and numpy sorting algorithms should be stable by default.

mdbenito: I don't see a problem with stability. You'd have to manually alter the private state of the object in order to mess things up.

src/pydvl/value/results.py
mdbenito: What happened to YAGNI? The moment we need to sort by stderr we can add this, ok?

src/pydvl/value/results.py
AnesBenmerzoug: By stability I meant that sorting multiple times should not affect the end result.

mdbenito: This is the case! the test is just an optimization to avoid resorting. It's a clear case of premature optimization. No ValuationResult is going to hold more than a few thousand items. But it's a trivial one. We can remove this in the future. #220

PR #232, opened 2022-12-19
Fix and improve map reduce

### LINKED ISSUE #165: Ideas for a change of interface to MapReduceJob
* A job contains all the info it needs to run. Therefore passing arguments to `__call__` doesn't make much sense.
* Instead, all arguments to the map function are passed upon construction and `put()` to the backend inside the constructor, thus hiding this implementation detail from the user of `MapReduceJob` .
* One such argument is necessarily the data to use, so it should not be part of `map_kwargs` but a standalone arg to the constructor of `MapReduceJob`
 
Additionally:
* we avoid duplicating arguments in `__call__` like `chunkify`, etc.
* we undo the typing workaround with `Any`
* `num_workers` in `ParalleConfig` should be renamed to `n_local_workers` with proper explanation of the difference with `n_jobs`

### LINKED ISSUE #193: Unexpected behaviour of MapReduceJob with multiple runs
The docstring states that if `n_runs > n_jobs`, then each job receives all data and a total of `n_runs` results are computed. However, in the method `map()`, the chunkify setting takes precedence and interacts with n_runs. If workload is split, one would expect `n_jobs=n_runs=2` to result in a total of two results, each computed in one worker which receives all data. But if `chunkify=True` is passed, then each job receives half the data so that one doesn't have two full runs. And if chukify is False, then each run contains two copies of the inputs, so that the function is called four times on all data.

AnesBenmerzoug:
### Description

This PR closes #193 and closes #165

### Changes

- Adds a `SequentialParallelBackend` class that runs all jobs in the current thread.
- Defines a `BaseParallelBackend` abstract class.
- Defines a `AbstractNoPublicConstructor` abstract metaclass to disallow directly instantiating parallel backends.
- Passes inputs to `MapReduceJob` at initialization.
- Removes `chunkify_inputs` parameter from `MapReduceJob`.
- Removes `n_runs` parameter from `MapReduceJob`.
- Calls the parallel backend's `put()` method for each generated chunk in `_chunkify()`.
- Skips chunkification in `MapReduceJob` if `n_runs` >= `n_jobs`.
- Use `singledispatch` and `singledispatchmethod` to handle different input types in `chunkify` and `_get_value`.
- Adds a `__repr__` method to `ValuationResult`.
- Calls `parallel_backend.put()` on the input to `MapReduceJob`'s `__call__` method if it is a sequence of `Utility` objects.
- Renames `ParallelConfig`'s `num_workers` attribute to `n_local_workers`.
- Small fixes to `MapReduceJob`'s docstring.

### Checklist

- [X] Updated Changelog


AnesBenmerzoug:
@mdbenito I have tried using `partial` to avoid using map_kwargs and reduce_kwargs but it didn't work with ray's `put` function. I checked their code and it does different things depending on the passed input.

src/pydvl/value/shapley/montecarlo.py
```diff
@@ -234,10 +231,13 @@ def permutation_montecarlo_shapley(
...
         reduce_kwargs=dict(axis=0),
         config=config,
-        chunkify_inputs=False,
         n_jobs=n_jobs,
     )
-    full_results = map_reduce_job(u_id)[0]
+    if n_jobs == 1:
+        input_ = u
+    else:
+        input_ = tuple(repeat(u, times=n_jobs))
```
mdbenito, line 239: Why the case handling? a tuple of `(u,)` should also be ok. As a matter of fact, I'd expect a map-reduce op to always take a sequence and never a single item.

src/pydvl/utils/parallel/map_reduce.py
```diff
@@ -175,24 +158,21 @@ def __init__(
 
     def __call__(
         self,
-        inputs: Union[Collection[T], Any],
-        *,
-        n_jobs: Optional[int] = None,
-        n_runs: Optional[int] = None,
-        chunkify_inputs: Optional[bool] = None,
+        inputs: Union[Sequence[T], Sequence[Utility], Utility],
```
mdbenito, line 161: Is there a reason not to pass the inputs to the constructor instead? IMO, a "Job" should contain all information required to run it, and any `call` or `run` method wouldn't take any arguments.

Also, see my other comment about accepting non-sequence objects. At first sight, it doesn't make much sense to accept single items for a parallel job and it doesn't take any effort for the user of the class to wrap one in a tuple in case it's needed

src/pydvl/utils/parallel/map_reduce.py
```diff
@@ -175,24 +158,21 @@ def __init__(
...
-        if chunkify_inputs is not None:
-            self.chunkify_inputs = chunkify_inputs
-
-        map_results = self.map(inputs)
+        if isinstance(inputs, Utility):
+            inputs_ = self.parallel_backend.put(inputs)
+        elif len(inputs) > 0 and isinstance(inputs[0], Utility):
+            inputs_ = [self.parallel_backend.put(x) for x in inputs]
+        else:
+            inputs_ = inputs
```
mdbenito, line 168: Why can only `put()` utility objects? Shouldn't this be the same for any type? I'd expect to have to put heavy np arrays to the store as well, right?

src/pydvl/utils/parallel/backend.py
```diff
@@ -15,10 +17,89 @@
...
+    @abstractmethod
+    def effective_n_jobs(self, n_jobs: Optional[int]) -> int:
+        ...
+
+    def __repr__(self) -> str:
+        return f"<{self.__class__.__name__}: {self.config}>"
+
+
+class SequentialParallelBackend(BaseParallelBackend):
+    """Class used to run jobs sequentiall and locally. It shouldn't
```
mdbenito, line 53:
```suggestion
    """Class used to run jobs sequentially and locally. It shouldn't
```

src/pydvl/utils/parallel/backend.py
```diff
@@ -15,10 +17,89 @@
...
+    be initialized directly. You should instead call `init_parallel_backend`.
+
+    :param config: instance of :class:`~pydvl.utils.config.ParallelConfig` with number of cpus
+
+    :Example:
+
+    >>> from pydvl.utils.parallel.backend import SequentialParallelBackend
+    >>> from pydvl.utils.config import ParallelConfig
+    >>> config = ParallelConfig(backend="sequential")
+    >>> parallel_backend = SequentialParallelBackend(config)
```
mdbenito, line 63: This feels a bit awkward: first I select the backend in a generic config class, then I need to choose the right class manually and pass it the config object with the backend. Since backends are not supposed to be instantiated manually, maybe it is not a big deal, but then the example should illustrate `init_parallel_backend()` instead of direct construction (which wouldn't work if we disable init with a metaclass as suggested above)

src/pydvl/utils/parallel/backend.py
```diff
@@ -15,10 +17,89 @@
...
+        config_dict = asdict(config)
+        config_dict.pop("backend")
+        config_dict.pop("address")
+        config_dict["num_cpus"] = config_dict.pop("n_local_workers")
+        self.config = config_dict
+
+    def get(self, v: Any, *args, **kwargs):
+        return v
+
+    def put(self, v: Any, *args, **kwargs) -> Any:
```
mdbenito, line 79: Shouldn't this store `v` in e.g. a `dict`, and return an id so that `get()` works on it? This would probably simplify things later (avoid handling different cases depending on the backend), right?

src/pydvl/utils/numeric.py
```diff
@@ -252,7 +253,7 @@ def top_k_value_accuracy(y_true: "NDArray", y_pred: "NDArray", k: int = 3) -> fl
 
 
 def sigmoid(x: float) -> float:
-    return float(1 / (1 + np.exp(-x)))
+    return expit(x).item()
```
mdbenito, line 256: We depend on scipy for spearman correlation and conjugate gradient, so in principle it's not a problem to add another place where we depend on it. But at the same time, we are unnecessarily increasing the difficulty of getting rid of the dependency at some point (e.g. if we find a better / more suitable CG and throw away the plots with spearman). Since scipy actually does `1/(1+np.exp(-x))` itself (according to the docs), I'd suggest to revert this change.

src/pydvl/utils/parallel/backend.py
```diff
@@ -15,10 +17,89 @@
...
+    def effective_n_jobs(self, n_jobs: Optional[int]) -> int:
+        ...
+
+    def __repr__(self) -> str:
+        return f"<{self.__class__.__name__}: {self.config}>"
+
+
+class SequentialParallelBackend(BaseParallelBackend):
+    """Class used to run jobs sequentiall and locally. It shouldn't
+    be initialized directly. You should instead call `init_parallel_backend`.
```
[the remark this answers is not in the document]
AnesBenmerzoug: That's a really good suggestion, thanks!

src/pydvl/utils/parallel/backend.py
AnesBenmerzoug: This sequential class is just a dummy class that returns the same object. It's actually an error that put returns nothing here, I have fixed it locally and will push it when I'm done with the current round of fixing the type hints.

src/pydvl/utils/parallel/map_reduce.py
AnesBenmerzoug: I agree, but I couldn't find a way to easily do this.

I have working on some changes to this locally while also fixing type hints. I will push them and ask for a feedback again.

src/pydvl/utils/numeric.py
mdbenito: It's more of a general philosophy to reduce dependencies than an optimisation but ok, I'll take those extra 7µs 😄

src/pydvl/utils/parallel/map_reduce.py
```diff
@@ -158,6 +168,11 @@ def __init__(
         else:
             self.max_parallel_tasks = max_parallel_tasks
 
+        if isinstance(inputs, Sequence):
+            self.inputs_: Union[SequenceType[T], "ObjectRef[T]"] = inputs
+        else:
+            self.inputs_ = self.parallel_backend.put(inputs)
+
```
mdbenito, line 175: I'm not sure I understand this logic: If the input is an ndarray, then we do _not_ `put()` it to the backend?

I know I keep going in circles around this question, but why can't we just `put` **all** inputs to the backend? If I'm not mireading, it makes for a cumbersome API, having to instantiate the parallel backend and `put()` the data to it before creating a parallel job.

Instead I'd expect `MapReduceJob` to know to `put()` the inputs and any arguments to the backend (except possibly primitive types, enums, strings, and a couple others)

src/pydvl/value/shapley/montecarlo.py
```diff
@@ -370,14 +366,14 @@ def combinatorial_montecarlo_shapley(
```
mdbenito, line 366: The two lines above (github wouldn't let me comment on those) are the kind of boilerplate that I think we should avoid, that I refer to in my previous comment. Is there a problem with `put()`ing all `map_kwargs` (which are not ints, bools, etc.)?

src/pydvl/utils/parallel/map_reduce.py
```diff
@@ -264,36 +269,55 @@ def _backpressure(
             n_finished += len(finished_jobs)
         return n_finished
 
+    @singledispatchmethod
+    @staticmethod
+    def _chunkify(data: Any, n_chunks: int):
+        raise NotImplementedError(
+            f"_chunkify does not support data of type {type(data)}"
+        )
```
mdbenito, line 277: Much as I like to see overloading instead of ugly elif switches, wouldn't this cause a problem with inheritance? If we wanted to add a new MapReduceJob which chunkifies differently, it's unclear where to register the new method and how it affects other inheriting classes, see [this discussion](https://stackoverflow.com/questions/68190543/how-to-use-singledispatchmethod-with-inheritance-class). OTOH, this is perhaps not that much of an issue, at least right now

mdbenito (commented):
This is looking nice :) I still have my previous question about putting everything to the backend, though. And a few comments here and there

src/pydvl/utils/parallel/map_reduce.py
AnesBenmerzoug: The issue with using `put()` on such objects that are meant to be split across jobs is that we would either have to call `get()` on them inside chunkify in order to split them or make `_chunkify()` a remote function just like the others and I think it would just make things harder to debug.

src/pydvl/utils/parallel/map_reduce.py
AnesBenmerzoug: That's something that we don't need right now and we if we do eventually need the suggestions in that discussion indicate that it is indeed possible without too much effort.

src/pydvl/value/shapley/montecarlo.py
AnesBenmerzoug: Actually no. using `put()` on all `map_kwargs` and `reduce_kwargs` should not cause an issue. I will try that now.

AnesBenmerzoug: Done.

src/pydvl/utils/parallel/map_reduce.py
mdbenito: But the "solutions" in that thread are not acceptable IMO. If we leave the singledispatch then we are consciously disallowing extension of MapReduce by users.

AnesBenmerzoug: I don't agree with allowing users to extend MapReduceJob since it is meant to be transparent to the users. Still I will remove it.

src/pydvl/utils/parallel/map_reduce.py
mdbenito: Hmm... as it is, it is bit confusing as an interface. Either we put all arguments, or we don't, because this separate handling makes bogus usage rather likely and also hard to debug: if one doesn't put() objects they are automagically copied to every remote function, right? If one doesn't know this, things will be slow and it will be hard to know why.

What about calling `put` inside `chunkify` for each chunk separately? the chunks are computed in the main process, right? Then each process would only get the chunk it needs.

AnesBenmerzoug: The problem with your suggestion is that unless we keep track of the `ObjectRef` instances for each chunk across calls to `_chunkify()` then we won't be gaining anything from doing it that way.

AnesBenmerzoug: What we could do instead is call `put()` on the passed input regardless of what it is and wrap `_chunkify()` the same way we do for `map_func` and `reduce_func`.

This would also open the possibility to have a `chunk_func` argument that would allow the possibility to change the chunkification without having to create a new subclass.

mdbenito: I'm not sure I understand the first objection, since chunkify is called by `map()` once. If you mean for repeated calls to map, then keeping track as you say should not be a problem at all, maybe even a simple `@lru_cache` can do the trick in a first approximation. If this sounds doable, we could try it, otherwise we can go for your suggestion and leave a comment for us to avoid copying all data to all workers.

As to the idea of custom chunkification, that sounds interesting, maybe we can create an issue for that too

AnesBenmerzoug:
@mdbenito I have made the changes we talked about:

- removed `n_runs` from `MapReduceJobs`
- call `put()` on each chunk inside `_chunkify()`

I also added two tests for `MapReduceJobs` and fixed a bug when passing numpy arrays to `_chunkify()`.

src/pydvl/utils/parallel/map_reduce.py
```diff
@@ -1,302 +1,301 @@
...
+                # The difference is that it only uses built-in functions
+                # and does not convert the input data to an array
+                chunk_size, remainder = divmod(n, n_chunks)
+                chunk_indices = tuple(
+                    accumulate(
+                        [0]
+                        + remainder * [chunk_size + 1]
+                        + (n_chunks - remainder) * [chunk_size]
+                    )
+                )
```
mdbenito, line 282:
```suggestion
        if n_chunks == 1:
            data_id = self.parallel_backend.put(data)
            return [data_id]

        try:
            # This is used as a check to determine whether data is iterable or not
            # if it's the former, then the value will be used to determine the chunk indices.
            n = len(data)
        except TypeError:
            data_id = self.parallel_backend.put(data)
            return list(repeat(data_id, times=n_chunks))
        else:
            # This is very much inspired by numpy's array_split function
            # The difference is that it only uses built-in functions
            # and does not convert the input data to an array
            chunk_size, remainder = divmod(n, n_chunks)
            chunk_indices = tuple(
                accumulate(
                    [0]
                    + remainder * [chunk_size + 1]
                    + (n_chunks - remainder) * [chunk_size]
                )
            )
```

src/pydvl/utils/parallel/map_reduce.py
```diff
@@ -1,302 +1,301 @@
...
+
+                for start_index, end_index in zip(
+                    chunk_indices[:-1], chunk_indices[1:]
+                ):
+                    if start_index >= end_index:
+                        break
+                    chunk_id = self.parallel_backend.put(data[start_index:end_index])
+                    chunks.append(chunk_id)
+
+                return chunks
```
mdbenito, line 294:
```suggestion
            chunks = []

            for start_index, end_index in zip(chunk_indices[:-1], chunk_indices[1:]):
                if start_index >= end_index:
                    break
                chunk_id = self.parallel_backend.put(data[start_index:end_index])
                chunks.append(chunk_id)

            return chunks
```

src/pydvl/value/shapley/naive.py
```diff
@@ -1,17 +1,12 @@
 import math
 import warnings
 from itertools import permutations
+from typing import List
 
 import numpy as np
+from numpy.typing import NDArray
```
mdbenito, line 7: Why don't we always directly import like this instead of using the if TYPING guard and then quoting all types? (that's a bit of a PITA, tbh)

mdbenito (approved):
I think this looks rather nice now :) I suggested a couple of cosmetic changes, ditching redundant elses, but other than that I'm 100% fine with merging. Go ahead!

src/pydvl/value/shapley/naive.py
AnesBenmerzoug: I think it was a leftover from when we didn't pin the minimum numpy version to 1.20.

The [typing](https://numpy.org/devdocs/reference/typing.html) module was added to numpy in version 1.20.

So, yes I think we should just import it directly.

src/pydvl/utils/parallel/map_reduce.py
```diff
@@ -1,302 +1,301 @@
...
-        self.map_kwargs = map_kwargs
-        self.reduce_kwargs = reduce_kwargs
-
-        if self.map_kwargs is None:
+        if map_kwargs is None:
             self.map_kwargs = dict()
+        else:
+            self.map_kwargs = {
+                k: self.parallel_backend.put(v) for k, v in map_kwargs.items()
+            }
```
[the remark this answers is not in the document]
AnesBenmerzoug: I think this makes it harder to read since this is not a common pattern.

PR #235, opened 2022-12-27
Fix/cnn influence

mdbenito:
@Xuzzo This is my review to your PR. I have made several changes, some of them possibly not ideal, but I couldn't devote much time to it and given the urgency decided to do something rather than nothing.

Here are a few comments:

Notebook:

* requirements are missing (torch, torchvision, datasets, PIL). These should be in an additional requirements file somewhere
* operations on the datasets are extremely slow (the first cells after loading, e.g. the info() and the types) on my laptop.
* I wrote a trivial class for model persistence, removing the variable `run_model_trainings`. As mentioned, we should strive for as little boilerplate as possible in documentation.
* RuntimeError with model saving (solved using `os.makedirs`)
* Training runs for fixed number of epochs without a stopping criterion seems sloppy.
* The claim that `compute_influences` takes a loss "which typically is the training loss, but not necessarily" cannot be correct: the influence is wrt. the loss used to train.
* We should reserve the prefix `get_...` for getter functions (better yet, ditch the redundant prefix). In particular, functions that compute are not getters
* Rephrased a bit some of the appendix, but didn't really look into it.

torch_wrappers.py:

* `InternalDataset`: the name is imprecise and misleading. It seems to be anadapter for torch's dataset to comply with pyDVL's interface. Or is it the other way around? This should be explicit and inherit where adequate. If necessary define a protocol.
* The import guard `_TORCH_INSTALLED` fails with `InternalDataset`, which requires torch installed to be evaluated by the interpreter (you can try running the notebook in an environment without torch and see how the error is completely confusing)

I made more changes here and there. Again: not the best, but hopefully helpful...


Xuzzo (approved):
Thanks for the help. A few comments on your comments:

Notebook:
- "requirements are missing": in theory, tox would provide the env. In #221 we discussed including a notebook req. I guess I can add it now
-  "operations on the datasets are extremely slow": this is strange. Apart for when I download the data, types and info are very fast on my laptop. Maybe low RAM?
- "I wrote a trivial class for model persistence": looks much better now. Thanks
- "RuntimeError with model saving": ok
- "Training runs for fixed number of epochs seems sloppy": in general I agree that it is sloppy, but in our case not searching for the absolute minimum of the loss (and rather get to a "good enough" optimum) makes a stronger point for influences. It shows that you can balance sloppy training with a bit of hessian regularisation and still have useful results.
- "The claim that compute_influences takes a loss": I agree that in theory they should be the same. But in practice, you can choose another one and balance it with hessian regularisation. Anyway, let's remove so we dont create confusion to the reader.
- "We should reserve the prefix get_... for getter functions": ok
- "Rephrased a bit some of the appendix": ok

torch_wrappers.py:
- "InternalDataset: the name is imprecise and misleading.": if I remember correctly, InternalDataset was introduced to make arrays compatible with DataLoaders. Will look into a better solution.
- "The import guard _TORCH_INSTALLED fails with InternalDataset": the error being misleading is a problem, but I dont think this is a failure of the import guard. Anyway, will see if there is a better solution to InternalDataset.

Xuzzo:
I will go ahead and merge this. I will do the rest of the changes in the other branch.

PR #243, opened 2023-01-11
Improve Least Core

AnesBenmerzoug:
### Description

This PR uses cvxpy instead of scipy to solve the Least Core linear program and then uses the found
least core value to solve a quadratic program to find an egalitarian distribution of data values.

This is exactly what is done in the original paper.

### Changes

- Adds cvxpy as a dependency.
- Modifies the function used to solve the least core linear program to use it instead of scipy.
- Adds a new function to solve the egalitarian least core quadratic program to find the data values.

### Checklist

- [x] Updated Changelog


src/pydvl/value/least_core/_common.py
```diff
@@ -1,108 +1,167 @@
...
+                "Solver terminated early. Consider increasing the solver's "
+                "maximum number of iterations in options",
+                RuntimeWarning,
+            )
+        least_core_value = e.value.item()
+        # HACK: sometimes the returned least core value
+        # is negative but very close to 0
+        # to avoid any problems with the subsequent quadratic program
+        # we just set it to 0.0
+        if least_core_value < 0:
```
mdbenito, line 84: Shouldn't one: a) check with isclose() or similar (to avoid clipping large deviations, if that's possible) b) warn that this happens in the logs

src/pydvl/value/least_core/_common.py
```diff
@@ -1,108 +1,167 @@
...
+        coefficients of a linear equality constraint on ``x``.
+    :param b_eq: The equality constraint vector. Each element of ``A_eq @ x`` must equal
+        the corresponding element of ``b_eq``.
+    :param A_lb: The inequality constraint matrix. Each row of ``A_lb`` specifies the
+        coefficients of a linear inequality constraint on ``x``.
+    :param b_lb: The inequality constraint vector. Each element represents a
+        lower bound on the corresponding value of ``A_lb @ x``.
+    :param options: Keyword arguments that will be used to select a solver
+        and to configure it. Refer to the following page for all possible options:
+        https://www.cvxpy.org/tutorial/advanced/index.html#setting-solver-options
```
mdbenito, line 50:
```suggestion
        and to configure it. For all possible options, refer to `cvxpy's documentation
        <https://www.cvxpy.org/tutorial/advanced/index.html#setting-solver-options>`_
```

src/pydvl/value/least_core/_common.py
```diff
@@ -1,108 +1,167 @@
...
+            RuntimeWarning,
+        )
+    return None, None
 
-    logger.debug(f"{result=}")
 
-    if result.success:
-        return np.asarray(result.x)
+def _solve_egalitarian_least_core_quadratic_program(
+    least_core_value: float,
```
mdbenito, line 97: Value is already quite overloaded and the paper uses "subsidy", which makes sense
```suggestion
    least_core_subsidy: float,
```

src/pydvl/value/least_core/_common.py
```diff
@@ -1,108 +1,167 @@
...
+    :param b_lb: The inequality constraint vector. Each element represents a
+        lower bound on the corresponding value of ``A_lb @ x``.
+    :param options: Keyword arguments that will be used to select a solver
+        and to configure it. Refer to the following page for all possible options:
+        https://www.cvxpy.org/tutorial/advanced/index.html#setting-solver-options
+    """
+    logger.debug(f"Solving quadratic program : {A_eq=}, {b_eq=}, {A_lb=}, {b_lb=}")
+
+    if least_core_value < 0:
+        raise ValueError("Passed least core value should be positive.")
```
mdbenito, line 135:
```suggestion
    if least_core_subsidy < 0:
        raise ValueError("The least core subsidy must be non-negative.")
```

src/pydvl/value/least_core/_common.py
```diff
@@ -1,108 +1,167 @@
...
+
+    if problem.status in cp.settings.SOLUTION_PRESENT:
+        logger.debug("Problem was solved")
+        if problem.status in [cp.settings.OPTIMAL_INACCURATE, cp.settings.USER_LIMIT]:
+            warnings.warn(
+                "Solver terminated early. Consider increasing the solver's "
+                "maximum number of iterations in options",
+                RuntimeWarning,
+            )
+        least_core_value = e.value.item()
```
mdbenito, line 79: See my comment below, value is already quite overloaded and the paper uses "subsidy"
```suggestion
        subsidy = e.value.item()
```

src/pydvl/value/least_core/_common.py
```diff
@@ -1,108 +1,167 @@
...
+        \mbox{such that} \ & A_{eq} x = b_{eq}, \\
+        & A_{lb} x + e \ge b_{lb},\\
+        & A_{eq} x = b_{eq},\\
+        & x in \mathcal{R}^n , \\
+        & e \text{ is a constant.}
+
+     where :math:`x` is a vector of decision variables; ,
+    :math:`b_{ub}`, :math:`b_{eq}`, :math:`l`, and :math:`u` are vectors; and
+    :math:`A_{ub}` and :math:`A_{eq}` are matrices.
+
```
mdbenito, line 119:
```suggestion

    :param least_core_subsidy: Minimal subsidy returned by :func:`_solve_least_core_linear_program`
```

src/pydvl/value/least_core/_common.py
```diff
@@ -1,108 +1,167 @@
...
+        )
+    return None, None
 
-    logger.debug(f"{result=}")
 
-    if result.success:
-        return np.asarray(result.x)
+def _solve_egalitarian_least_core_quadratic_program(
+    least_core_value: float,
+    n_variables: int,
```
mdbenito, line 98: Can't `n_variables` be inferred from `A_eq.shape[1]`? I'd remove the arg.

tests/value/__init__.py
```diff
@@ -59,11 +63,19 @@ def check_values(
...
+    :param extra_values_names: Sequence of names of extra values that should
+        also be compared.
     """
     values.sort()
     exact_values.sort()
 
     assert np.allclose(values.values, exact_values.values, rtol=rtol, atol=atol)
+    if extra_values_names is None:
+        return
+    for name in extra_values_names:
```
mdbenito, line 75: Maybe default the attribute to an empty tuple in order to avoid the None check

tests/value/least_core/conftest.py
```diff
@@ -56,11 +56,14 @@ def miner_utility(request) -> Tuple[Utility, ValuationResult]:
     u = MinerUtility(n_miners=n_miners)
     if n_miners % 2 == 0:
         exact_values = np.array([0.5] * n_miners)
+        least_core_value = 0.0
```
mdbenito, line 59: subsidy

mdbenito (commented):
Some comments about names and a couple minor suggestions. I haven't reviewed the notebook, only the first cell, where I think the sentence

> Whereas the Shapley values define a fair way to distribute data values amongst all data points when the entire training set is used. But they do not consider the question of stability: Would some data points get a better payoff by using smaller subsets?

Could be changed a bit, e.g. 

> Shapley values define a fair way of distributing the worth of the whole training set when every data point is part of it. But they do not consider the question of stability of subsets: Could some data points obtain a higher payoff if they formed smaller subsets? It is argued that this might be relevant if data providers are paid based on data value, since Shapley values can incentivise them not to contribute their data to the "grand coalition", but instead try to form smaller ones. Whether or not this is of actual practical relevance is debatable, but in any case, the least core is an alternative tool available for any task of Data Valuation

src/pydvl/value/least_core/_common.py
AnesBenmerzoug: `isclose()` would not be the right solution since it would check for closeness regardless of sign. This simple comparison should work. Large deviations should not exist since in that case it would mean that the solver failed to find a solution. I will add the warning though as it make sense to not hide this from the user.

src/pydvl/value/least_core/_common.py
AnesBenmerzoug: You're right, it's overloaded. The paper also uses the term "deficit" and comparing the definition of both terms in a dictionary I would lean towards using "deficit" rather than subsidy.

* **deficit**: the amount by which something, especially a sum of money, is too small.
* **subsidy**: a sum of money granted to support an undertaking held to be in the public interest.

AnesBenmerzoug: I just double-checked and it seems to be also just called **least core value** in the literature: https://www.sciencedirect.com/science/article/pii/S1572528613000042#br000100

mdbenito: Still, I'd avoid the overloading. For us values are assigned to samples.

src/pydvl/value/least_core/_common.py
AnesBenmerzoug: Done.

tests/value/__init__.py
AnesBenmerzoug: Done.

src/pydvl/value/least_core/_common.py
AnesBenmerzoug: Hmmm, I see. your reasoning makes sense. I'll change it to subsidy.

mdbenito (approved):

PR #247, opened 2023-01-15
Fix using remote ray actor with external cluster

AnesBenmerzoug:
### Description

This PR fixes a bug when instantiating `RayActorWrapper` on an Actor that's present on an external cluster (i.e. one that was not started with `ray.init()`.

Additionally, this modifies the `Utility` class' handling of warnings so that when using ray with small subsets
we don't end up seeing the same warning message over and over again across workers (e.g. using Naive Bayes with TMC).

### Changes

- Instantiates the worker and coordinator actors directly inside `RayActorWrapper`.
- Tests Shapley methods with different parallel backends.
- Raises an error when attempting use TMC with the sequential backend.
- Adds a warning to TMC's docstring.
- Catch all warnings in Utility when show_warnings is set to False

### Checklist

- [x] Updated Changelog

### DIFF
--- a/src/pydvl/utils/utility.py
+++ b/src/pydvl/utils/utility.py
@@ -267,14 +267,15 @@ class MinerUtility(Utility):
 
     Consider a group of n miners, who have discovered large bars of gold.
 
-    If two miners can carry one piece of gold, then the payoff of a coalition S is:
+    If two miners can carry one piece of gold,
+    then the payoff of a coalition $S$ is:
 
-    $$
-    v(S) = \left\{\begin{matrix}
-        |S|/2 & \text{, if} & |S| \text{ is even} \\
-        (|S| - 1)/2 & \text{, if} & |S| \text{ is odd}
-    \end{matrix}\right.
-    $$
+    $${
+    v(S) = \left\{\begin{array}{lll}
+    | S | / 2 & \text{, if} & | S | \text{ is even} \\
+    ( | S | - 1)/2 & \text{, if} & | S |  \text{ is odd}
+    \end{array}\right.
+    }$$
 
     If there are more than two miners and there is an even number of miners,
     then the core consists of the single payoff where each miner gets 1/2.
--- a/src/pydvl/utils/utility.py
+++ b/src/pydvl/utils/utility.py
@@ -41,9 +41,19 @@ class Utility:
     the computation of :ref:`Shapley values<Shapley>` and
     :ref:`Least Core values<Least Core>`.
 
-    Since evaluating the scoring function requires retraining the model, this
-    class wraps it and caches the results of each execution. Caching is
-    available both locally and across nodes, but must always be enabled for your
+    The Utility expect the model to fulfill
+    the :class:`pydvl.utils.types.SupervisedModel` interface
+    i.e. to have a `fit()`, `predict()`, and `score()` methods.
+
+    When calling the utility, the model will be
+    `cloned <https://scikit-learn.org/stable/modules/generated/sklearn.base.clone.html>`_
+    if it is a Sci-Kit Learn model, otherwise a copy is created using `deepcopy()`
+    from the builtin `copy <https://docs.python.org/3/library/copy.html>`_ module.
+
+    Since evaluating the scoring function requires retraining the model
+    and that can be time consuming, this class wraps it and caches
+    the results of each execution. Caching is available both locally
+    and across nodes, but must always be enabled for your
     project first, see :ref:`how to set up the cache<caching setup>`.
 
     :param model: Any supervised model. Typical choices can be found at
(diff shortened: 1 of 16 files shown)


src/pydvl/utils/utility.py
```diff
@@ -145,21 +149,30 @@ def _utility(self, indices: FrozenSet) -> float:
...
+                warnings.simplefilter("ignore")
+            try:
+                # Clone the model to avoid the possibility
+                # of reusing a fitted estimator
+                try:
+                    model = clone(self.model)
+                except TypeError:
+                    # This happens if the passed model is not an sklearn model
+                    # In this case, we just make a deepcopy of the model.
+                    model = clone(self.model, safe=False)
```
mdbenito, line 163: I think the cloning could be optional (defaulting to True). In any case, we should document that it is done.

src/pydvl/utils/utility.py
```diff
@@ -33,8 +37,9 @@ class Utility:
     """Convenience wrapper with configurable memoization of the scoring function.
 
     An instance of `Utility` holds the triple of model, dataset and scoring
-    function which determines the value of data points. This is mostly used for
-    the computation of :ref:`Shapley values<data valuation>`.
+    function which determines the value of data points. This is only used for
+    the computation of :ref:`Shapley values<Shapley>` and
+    :ref:`Least Core values<Least Core>`.
```
mdbenito, line 42: "mostly" was future-proofing the docs: since we will soon add semivalues, we would have to remember to change the docstring is we use "only"

src/pydvl/utils/utility.py
```diff
@@ -259,3 +272,112 @@ def __call__(self, indices: Iterable[int]) -> float:
...
+        y = np.zeros_like(x)
+
+        self.data = Dataset(x_train=x, y_train=y, x_test=x, y_test=y)
+
+    def __call__(self, indices: Iterable[int]) -> float:
+        n = len(tuple(indices))
+        if n % 2 == 0:
+            return n / 2
+        else:
+            return (n - 1) / 2
```
mdbenito, line 318:
```suggestion
        return len(tuple(indices)) // 2
```

src/pydvl/utils/utility.py
```diff
@@ -259,3 +272,112 @@ def __call__(self, indices: Iterable[int]) -> float:
...
+    def exact_least_core_values(self) -> Tuple[NDArray[np.float_], float]:
+        if self.left == self.right:
+            values = np.array([0.5] * (self.left + self.right))
+            subsidy = 0.0
+        elif self.left < self.right:
+            values = np.array([1.0] * self.left + [0.0] * self.right)
+            subsidy = 0.0
+        else:
+            values = np.array([0.0] * self.left + [1.0] * self.right)
+            subsidy = 0.0
```
mdbenito, line 382:
```suggestion
        subsidy = 0.0
        if self.left == self.right:
            values = np.array([0.5] * (self.left + self.right))
        elif self.left < self.right:
            values = np.array([1.0] * self.left + [0.0] * self.right)
        else:
            values = np.array([0.0] * self.left + [1.0] * self.right)
```

src/pydvl/value/shapley/montecarlo.py
```diff
@@ -95,6 +95,10 @@ def truncated_montecarlo_shapley(
        criterion detailed in
        :meth:`~pydvl.value.shapley.actor.ShapleyCoordinator.check_done`.
 
+    .. warning::
+
+       This function does not work the sequential parallel backend.
```
mdbenito, line 100:
```suggestion
       This function does not work with the sequential parallel backend.
```

mdbenito (commented):
Looks good. I left my usual pesky suggestions and small comments :)

src/pydvl/utils/utility.py
AnesBenmerzoug: To be honest, I find the version with the `if else` more readable.

src/pydvl/utils/utility.py
AnesBenmerzoug: Okay, I will change it to mostly.

src/pydvl/utils/utility.py
AnesBenmerzoug: True, it should be documented. Don't you think adding yet another argument to the already packed `Utility` constructor is too much?

AnesBenmerzoug: Done.

tests/utils/test_utility.py
```diff
@@ -70,3 +105,5 @@ def test_different_cache(linear_dataset, memcache_client_config):
     )
 
     assert u1.signature != u2.signature
+    assert u1.signature == u1.signature
+    assert u2.signature == u2.signature
```
mdbenito, line 109: Is this correct?

src/pydvl/utils/utility.py
```diff
@@ -91,20 +106,19 @@ def __init__(
         show_warnings: bool = False,
         enable_cache: bool = False,
         cache_options: Optional[MemcachedConfig] = None,
+        clone_before_fit: bool = True,
     ):
-        self.model = model
+        self.model = self._clone_model(model)
```
mdbenito, line 111: We should document this too: if a clone is expensive, the user must know this will happen

mdbenito (approved):
I think there is a little mistake in the tests but I'm going to merge and fix it in my PR if that's ok bc I need to merge in these changes tonight.

tests/utils/test_utility.py
AnesBenmerzoug: Yes. I just wanted to be extra sure since it's a property.
Right now it's redundant though.

src/pydvl/utils/utility.py
AnesBenmerzoug: You're right. Sorry for not doing it.

PR #248, opened 2023-01-17
TMCS stopping, 1 of ...

### LINKED ISSUE #241: Implement stopping criterion from TMC paper
(instead of stddev)

mdbenito:
### Description

This PR is a first step towards implementing general stopping criteria

### Changes

- Adds operations on ValuationResults and (Valuation)Status
- Renames ValuationStatus to Status (since it's independent of the algorithm)
- Moves a couple of classes around to avoid circular imports
- Does some cleanup

### Checklist

- [x] Updated Changelog


src/pydvl/value/shapley/montecarlo.py
```diff
@@ -40,23 +41,21 @@
...
-    ParallelConfig,
-    Utility,
-    get_running_avg_variance,
-    init_parallel_backend,
-    maybe_progress,
-    random_powerset,
-)
-from pydvl.value import ValuationResult, ValuationStatus
-
+from ...utils.config import ParallelConfig
```
AnesBenmerzoug, line 44: Seeing this makes me realize that absolute imports are more readable.

src/pydvl/value/shapley/types.py
```diff
@@ -0,0 +1,35 @@
+from enum import Enum
+from typing import Protocol
+
+import numpy as np
+from numpy._typing import NDArray
+
+
+class ShapleyMode(str, Enum):
```
AnesBenmerzoug, line 8: Moving this here makes it a bit more cumbersome to import and use for the users.

AnesBenmerzoug (approved):
Looks good to me.

I left 2 comments but feel free to ignore them for now.

src/pydvl/value/shapley/montecarlo.py
mdbenito: Agreed, I think pycharm suggested it and I kept using them, thinking "maybe we'd want to rearrange stuff later on", but I'll revert to absolute ones in the 2nd PR

src/pydvl/value/shapley/types.py
mdbenito: I think I moved it to avoid some circular import. I'll try to move it back in the next PR.

PR #250, opened 2023-01-19
Generalised stopping criteria, truncation policies and improved data structures

### LINKED ISSUE #212: Permutation MC wastes memory
by keeping the history of all values and then computing mean and stderr. Instead, we should do as in the other methods and use a running average

### LINKED ISSUE #219: Add key argument to ValuationResult.sort
It would be nice to offer this interface instead to allow using the std in the comparison, for example
```suggestion
    def sort(self, *, key=None, reverse=False) -> None:
```
But this would mean that we cannot use argsort, which would make this slower.

_Originally posted by @AnesBenmerzoug in https://github.com/appliedAI-Initiative/pyDVL/pull/214#discussion_r1040666812_

### LINKED ISSUE #220: Simplify ValuationResult.sort
I think it would be better if we make this method completely stateless. It just adds complexity to the code without much benefit.

_Originally posted by @AnesBenmerzoug in https://github.com/appliedAI-Initiative/pyDVL/pull/214#discussion_r1040636604_

### LINKED ISSUE #241: Implement stopping criterion from TMC paper

mdbenito:
### Description

This PR closes #241 

### Changes

- Adds `StoppingCriterion` for use with (some) Monte Carlo methods
- Implements min and max updates, max time, standard error checks, and Ghorbani's criterion
- Adds `TruncationPolicy` for use with TMCS (to reproduce all choices in the paper, including those in the appendix)
- Improves implementation of `ValuationResult` to allow adding of results with different index sets
- Removes `MonteCarloResult` in favour of `ValuationResult` everywhere
- Simplifies some code and removes stale one.
- Closes #219 with a string key instead of a lambda for simplicity (should be enough since the container is not generic)
- Closes #220 by removing state, which would have been too complex after adding support for different sort keys
- Closes #212 by using running moments in `ValuationResult`

### Checklist

- [x] Updated Changelog

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,7 +7,7 @@ repos:
       - id: black-jupyter
         language_version: python3
   - repo: https://github.com/PyCQA/isort
-    rev: 5.10.1
+    rev: 5.12.0
     hooks:
       - id: isort
   - repo: https://github.com/kynan/nbstripout
--- a/build_scripts/update_docs.py
+++ b/build_scripts/update_docs.py
@@ -24,9 +24,6 @@ def module_template(module_qualname: str):
    :undoc-members:
    
    ----
-   
-   Module members
-   ==============
  
 .. footbibliography::
 
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,5 @@
 black[jupyter] == 22.10.0
-isort == 5.10.1
+isort == 5.12.0
 jupyter
 mypy == 0.982
 nbconvert
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,13 +2,13 @@ black[jupyter] == 22.10.0
 isort == 5.12.0
 jupyter
 mypy == 0.982
-nbconvert
+nbconvert>=7.2.9
 nbstripout == 0.6.1
 bump2version
-pre-commit == 2.20.0
-pytest
+pre-commit==3.0.4
+pytest==7.2.1
 pytest-cov
-pytest-docker
+pytest-docker==0.12.0
 pytest-mock
 pytest-timeout
 ray[default] >= 0.8
--- a/src/pydvl/reporting/scores.py
+++ b/src/pydvl/reporting/scores.py
@@ -6,7 +6,7 @@ import numpy as np
 from numpy.typing import NDArray
 
 from pydvl.utils import Utility, maybe_progress
-from pydvl.value.results import ValuationResult
+from pydvl.value.result import ValuationResult
 
 __all__ = [
     "sort_values",
@@ -66,11 +66,7 @@ def compute_removal_score(
     # We sort in descending order if we want to remove the best values
     values.sort(reverse=remove_best)
 
-    for pct in maybe_progress(
-        percentages,
-        display=progress,
-        desc="Removal Scores",
-    ):
+    for pct in maybe_progress(percentages, display=progress, desc="Removal Scores"):
         n_removal = int(pct * len(u.data))
         indices = values.indices[n_removal:]
         score = u(indices)
--- a/src/pydvl/reporting/scores.py
+++ b/src/pydvl/reporting/scores.py
@@ -1,6 +1,4 @@
-from collections import OrderedDict
-from operator import itemgetter
-from typing import Dict, Iterable, Mapping, Sequence, TypeVar, Union
+from typing import Dict, Iterable, Union
 
 import numpy as np
 from numpy.typing import NDArray
@@ -8,36 +6,13 @@ from numpy.typing import NDArray
 from pydvl.utils import Utility, maybe_progress
 from pydvl.value.result import ValuationResult
 
-__all__ = [
-    "sort_values",
-    "sort_values_array",
-    "sort_values_history",
-    "compute_removal_score",
-]
-
-KT = TypeVar("KT")
-VT = TypeVar("VT")
-
-
-def sort_values_array(values: np.ndarray) -> Dict[int, "NDArray"]:
-    vals = np.mean(values, axis=1)
-    return OrderedDict(sorted(enumerate(vals), key=itemgetter(1)))
-
-
-def sort_values_history(values: Mapping[KT, Sequence[VT]]) -> Dict[KT, Sequence[VT]]:
-    """Sorts a dict of sample_id: [values] by the last item in each list."""
-    return OrderedDict(sorted(values.items(), key=itemgetter(1, -1)))
-
-
-def sort_values(values: Mapping[KT, VT]) -> Dict[KT, VT]:
-    """Sorts a dict of sample_id: value_float by value."""
-    return OrderedDict(sorted(values.items(), key=itemgetter(1)))
+__all__ = ["compute_removal_score"]
 
 
 def compute_removal_score(
     u: Utility,
     values: ValuationResult,
-    percentages: Union["NDArray", Iterable[float]],
+    percentages: Union[NDArray[np.float_], Iterable[float]],
     *,
     remove_best: bool = False,
     progress: bool = False,
--- a/src/pydvl/utils/caching.py
+++ b/src/pydvl/utils/caching.py
@@ -271,7 +271,7 @@ def memcached(
                     ):
                         new_value = fun(*args, **kwargs)
                         new_avg, new_var = running_moments(
-                            value, variance, cast(float, new_value), int(count)
+                            value, variance, int(count), cast(float, new_value)
                         )
                         result_dict["value"] = new_avg
                         result_dict["count"] = count + 1
--- a/src/pydvl/utils/numeric.py
+++ b/src/pydvl/utils/numeric.py
@@ -69,20 +69,22 @@ def num_samples_permutation_hoeffding(eps: float, delta: float, u_range: float) 
 
 
 def random_powerset(
-    s: NDArray[T], max_subsets: Optional[int] = None, q: float = 0.5
+    s: NDArray[T], n_samples: Optional[int] = None, q: float = 0.5
 ) -> Generator[NDArray[T], None, None]:
     """Samples subsets from the power set of the argument, without
     pre-generating all subsets and in no order.
 
     See `powerset()` if you wish to deterministically generate all subsets.
 
-    To generate subsets, `len(s)` Bernoulli draws with probability `q` are drawn.
+    To generate subsets, `len(s)` Bernoulli draws with probability `q` are
+    drawn.
     The default value of `q = 0.5` provides a uniform distribution over the
     power set of `s`. Other choices can be used e.g. to implement
-    :func:`Owen sampling <pydvl.value.shapley.montecarlo.owen_sampling_shapley>`.
+    :func:`Owen sampling
+    <pydvl.value.shapley.montecarlo.owen_sampling_shapley>`.
 
     :param s: set to sample from
-    :param max_subsets: if set, stop the generator after this many steps.
+    :param n_samples: if set, stop the generator after this many steps.
         Defaults to `np.iinfo(np.int32).max`
     :param q: Sampling probability for elements. The default 0.5 yields a
         uniform distribution over the power set of s.
@@ -99,9 +101,9 @@ def random_powerset(
 
     rng = np.random.default_rng()
     total = 1
-    if max_subsets is None:
-        max_subsets = np.iinfo(np.int32).max
-    while total <= max_subsets:
+    if n_samples is None:
+        n_samples = np.iinfo(np.int32).max
+    while total <= n_samples:
         selection = rng.uniform(size=len(s)) > q
         subset = s[selection]
         yield subset
--- a/src/pydvl/utils/numeric.py
+++ b/src/pydvl/utils/numeric.py
@@ -230,8 +230,8 @@ def linear_regression_analytical_derivative_d_x_d_theta(
 def running_moments(
     previous_avg: FloatOrArray,
     previous_variance: FloatOrArray,
-    new_value: FloatOrArray,
     count: IntOrArray,
+    new_value: FloatOrArray,
 ) -> Tuple:  # [FloatOrArray, FloatOrArray]:
     """Uses Welford's algorithm to calculate the running average and variance of
      a set of numbers.
@@ -250,8 +250,8 @@ def running_moments(
 
     :param previous_avg: average value at previous step
     :param previous_variance: variance at previous step
-    :param new_value: new value in the series of numbers
     :param count: number of points seen so far
+    :param new_value: new value in the series of numbers
     :return: new_average, new_variance, calculated with the new count
     """
     # broadcasted operations seem not to be supported by mypy, so we ignore the type
--- a/src/pydvl/utils/parallel/actor.py
+++ b/src/pydvl/utils/parallel/actor.py
@@ -1,7 +1,7 @@
 import abc
 import inspect
 import logging
-from typing import Generic, List, TypeVar, Optional, Type, cast
+from typing import Generic, List, Optional, Type, TypeVar, cast
 
 from ..config import ParallelConfig
 from ..status import Status
--- a/src/pydvl/utils/parallel/actor.py
+++ b/src/pydvl/utils/parallel/actor.py
@@ -1,6 +1,7 @@
 import abc
 import inspect
 import logging
+from time import sleep
 from typing import Generic, List, Optional, Type, TypeVar, cast
 
 from ..config import ParallelConfig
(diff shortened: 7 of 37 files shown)


mdbenito:
@AnesBenmerzoug Latest pipeline didn't go through seemingly [because of memcache](https://github.com/appliedAI-Initiative/pyDVL/actions/runs/4186615549/jobs/7255478902#step:6:98). Have you seen this before? I have had some weirdness with the pytest-docker plugin locally (I think an updated version requires an additional fixture, but I'm not sure)

src/pydvl/reporting/scores.py
```diff
@@ -1,43 +1,18 @@
...
+from typing import Dict, Iterable, Union
 
 import numpy as np
 from numpy.typing import NDArray
 
 from pydvl.utils import Utility, maybe_progress
-from pydvl.value.results import ValuationResult
+from pydvl.value.result import ValuationResult
 
-__all__ = [
```
mdbenito, line 11: Unused code

requirements-dev.txt
```diff
@@ -1,14 +1,14 @@
...
+nbconvert>=7.2.9
 nbstripout == 0.6.1
 bump2version
-pre-commit == 2.20.0
-pytest
+pre-commit==3.0.4
+pytest==7.2.1
 pytest-cov
-pytest-docker
+pytest-docker==0.12.0
```
mdbenito, line 11: Had to pin: v1.0 and beyond require a new fixture `docker_compose_command`, but after adding it, docker wouldn't be able to start the container because of networking issues 🤷🏽

src/pydvl/utils/caching.py
```diff
@@ -271,7 +271,7 @@ def __call__(self, *args, **kwargs) -> T:
                     ):
                         new_value = fun(*args, **kwargs)
                         new_avg, new_var = running_moments(
-                            value, variance, cast(float, new_value), int(count)
```
mdbenito, line 274: Flipped the last to args for clarity

build_scripts/update_docs.py
```diff
@@ -24,9 +24,6 @@ def module_template(module_qualname: str):
    :undoc-members:
    
    ----
-
```
mdbenito, line 27: This was just visual noise in the docs

.pre-commit-config.yaml
```diff
@@ -7,7 +7,7 @@ repos:
       - id: black-jupyter
         language_version: python3
   - repo: https://github.com/PyCQA/isort
-    rev: 5.10.1
+    rev: 5.12.0
```
mdbenito, line 10: Solves some issue with poetry and blahblah... 😪

src/pydvl/utils/numeric.py
```diff
@@ -69,20 +69,22 @@ def num_samples_permutation_hoeffding(eps: float, delta: float, u_range: float)
 
 
 def random_powerset(
-    s: NDArray[T], max_subsets: Optional[int] = None, q: float = 0.5
+    s: NDArray[T], n_samples: Optional[int] = None, q: float = 0.5
```
mdbenito, line 72: Trying to be more consistent with arg names

src/pydvl/utils/parallel/actor.py
```diff
@@ -65,50 +67,45 @@ def wrapper(
...
 
-        :return: `True` if workers must terminate, `False` otherwise.
+        :return: ``True`` if workers must terminate, ``False`` otherwise.
         """
-        return self._is_done
+        return bool(self._status)
 
     @abc.abstractmethod
-    def get_results(self) -> Any:
+    def accumulate(self) -> Result:
```
mdbenito, line 102: More consistent naming

src/pydvl/utils/parallel/actor.py
```diff
@@ -117,25 +114,22 @@ class Worker(abc.ABC):
...
         worker_id: int,
         *,
-        progress: bool = False,
         update_period: int = 30,
     ):
         """A worker
 
         :param coordinator: worker results will be pushed to this coordinator
         :param worker_id: id used for reporting through maybe_progress
-        :param progress: set to True to report progress, else False
```
mdbenito, line 130: We don't support it yet, so best to drop the arg until we do.

src/pydvl/utils/parallel/map_reduce.py
```diff
@@ -70,8 +69,9 @@ def _(v: np.ndarray, *, timeout: Optional[float] = None) -> NDArray:
     return v
 
 
+# Careful to use list as hint. The dispatch does not work with typing generics
 @_get_value.register
-def _(v: Iterable, *, timeout: Optional[float] = None) -> List[Any]:
+def _(v: list, *, timeout: Optional[float] = None) -> List[Any]:
```
mdbenito, line 74: This was a bug: dispatch needs the actual type, not generics (at least in our python versions)

src/pydvl/value/__init__.py
```diff
@@ -5,7 +5,7 @@
 implemented here.
 """
 
-from .results import *  # isort: skip
+from .result import *  # isort: skip
```
mdbenito, line 8: More consistent naming

src/pydvl/value/least_core/naive.py
```diff
@@ -128,6 +117,5 @@ def exact_least_core(
         status=status,
         values=values,
         subsidy=subsidy,
-        stderr=None,
```
mdbenito, line 131: We use variances now, and it's also not required to init them

src/pydvl/value/result.py
```diff
@@ -0,0 +1,580 @@
...
+
+    @property
+    def stderr(self) -> Optional[float]:
+        """Standard error of the value."""
+        if self.variance is None or self.count is None:
+            return None
+        return float(np.sqrt(self.variance / self.count).item())
+
+
+class ValuationResult(collections.abc.Sequence):
```
mdbenito, line 86: Stupid git move and diffs don't go well together, sorry. Key points here are:
* sorting is simplified and hopefully indexing more consistent and pandas-like (and yes, I've finally realized that this object would be much simpler with a DataFrame as backend 🤦🏽)
* it's possible to sort by different keys, but I decided to deviate a bit from the standard interface for simplicity (sorry)
* default indexing of all attributes is based on data indices
* ipos-equivalent indexing is done with a dedicated method `get` (maybe I could rename that).
* Addition of results is also more stable and generalised to any data indices, making the `or` operator superfluous. This makes aggregating parallel results much simpler (just add them and it always works)

src/pydvl/value/shapley/__init__.py
```diff
@@ -23,15 +22,17 @@
     permutation_exact_shapley,
 )
 from pydvl.value.shapley.types import ShapleyMode
+from pydvl.value.stopping import MaxUpdates, StoppingCriterion
 
 __all__ = ["compute_shapley_values"]
 
 
 def compute_shapley_values(
```
mdbenito, line 30: There are still a few inconsistencies in this function, mostly wrt. stopping and extra kwargs. But it's ok-ish for now I think.

src/pydvl/value/shapley/actor.py
```diff
@@ -59,99 +59,42 @@ class ShapleyCoordinator(Coordinator):
...
-        of iterations have crossed the set thresholds.
-
-        If any of the thresholds have been reached, then calls to
-        :meth:`~Coordinator.is_done` return `True`.
-
-        :return: True if converged or reached max iterations.
+        if len(self.worker_results) == 0:
+            return ValuationResult.empty()
+
+        # FIXME: inefficient, possibly unstable
```
mdbenito, line 79: not so sure about this comment any more. I think it's fine

src/pydvl/value/shapley/actor.py
```diff
@@ -160,14 +103,16 @@ class ShapleyWorker(Worker):
     It should work.
     """
 
+    algorithm: str = "truncated_montecarlo_shapley"
```
mdbenito, line 106: Used for the valuation result

src/pydvl/value/shapley/actor.py
```diff
@@ -179,71 +124,48 @@ def __init__(
...
-        # Import here to avoid errors with circular imports
-        from .montecarlo import _permutation_montecarlo_marginals
+    def _compute_marginals(self) -> ValuationResult:
+        # Avoid circular imports
+        from .montecarlo import _permutation_montecarlo_shapley
 
-        return _permutation_montecarlo_marginals(self.u, max_permutations=1)[0]  # type: ignore
+        return _permutation_montecarlo_shapley(
+            self.u,
+            done=MaxChecks(1),
```
mdbenito, line 144: MaxChecks is a special policy that increments a counter on each call to check(), instead of looking at number of updates and so on.

src/pydvl/value/shapley/montecarlo.py
```diff
@@ -68,14 +63,149 @@ class MonteCarloResults(NamedTuple):
 ]
 
 
+class TruncationPolicy(abc.ABC):
```
mdbenito, line 66: I will move this and everything for TMCS to another module after this PR is merged

src/pydvl/value/shapley/montecarlo.py
```diff
@@ -86,123 +216,139 @@ def truncated_montecarlo_shapley(
     over all possible permutations of the index set, with a double stopping
     criterion.
 
-    .. warning::
```
mdbenito, line 89: Yay! finally gone!

src/pydvl/value/shapley/montecarlo.py
```diff
@@ -86,123 +216,139 @@ def truncated_montecarlo_shapley(
...
+    return coordinator.accumulate()
 
-    return ValuationResult(
-        algorithm="truncated_montecarlo_shapley",
-        status=coordinator.status(),
-        values=values,
-        stderr=stderr,
-        data_names=u.data.data_names,
-    )
+    # Something like this would be nicer, but it doesn't seem to be possible
```
mdbenito, line 286: I'll remove this comment, but I don't want to run the pipelines again just yet ... 😅

src/pydvl/value/shapley/montecarlo.py
```diff
@@ -86,123 +216,139 @@ def truncated_montecarlo_shapley(
...
-        column being the score of a different data point
+    :return: An object with the results
     """
-    n = len(u.data)
-    values = np.zeros(shape=(max_permutations, n))
-    pbar = maybe_progress(max_permutations, progress, position=job_id)
-    for iter_idx in pbar:
+    result = ValuationResult.empty(algorithm=algorithm_name, indices=u.data.indices)
+
+    pbar = tqdm(disable=not progress, position=job_id, total=100, unit="%")
```
mdbenito, line 326: Not using `maybe_progress`, but still ugly. We need a one-liner that updates a percentage

src/pydvl/value/shapley/montecarlo.py
```diff
@@ -216,117 +362,81 @@ def permutation_montecarlo_shapley(
...
 
-    map_reduce_job: MapReduceJob[Utility, "NDArray"] = MapReduceJob(
+    map_reduce_job: MapReduceJob[Utility, ValuationResult] = MapReduceJob(
         u,
-        map_func=_permutation_montecarlo_marginals,
-        reduce_func=np.concatenate,  # type: ignore
-        map_kwargs=dict(max_permutations=iterations_per_job, progress=progress),
-        reduce_kwargs=dict(axis=0),
+        map_func=_permutation_montecarlo_shapley,
+        reduce_func=lambda results: reduce(operator.add, results),
```
mdbenito, line 379: I find this much cleaner: all we need to do is add results together, and we get a valid result

src/pydvl/value/shapley/montecarlo.py
```diff
@@ -216,117 +362,81 @@ def permutation_montecarlo_shapley(
...
-    :param n_iterations: total number of subsets to sample.
+    :param done: Check on the results which decides when to stop sampling
+        subsets for an index.
     :param progress: Whether to display progress bars for each job.
     :param job_id: id to use for reporting progress
     :return: A tuple of ndarrays with estimated values and standard errors
     """
     n = len(u.data)
 
-    if len(np.unique(indices)) != len(indices):
```
mdbenito, line 273: Check done in ValuationResult

src/pydvl/value/shapley/montecarlo.py
```diff
@@ -216,117 +362,81 @@ def permutation_montecarlo_shapley(
...
-        raise ValueError("Repeated indices passed")
-
     # Correction coming from Monte Carlo integration so that the mean of the
     # marginals converges to the value: the uniform distribution over the
     # powerset of a set with n-1 elements has mass 2^{n-1} over each subset. The
     # additional factor n corresponds to the one in the Shapley definition
     correction = 2 ** (n - 1) / n
+    result = ValuationResult.empty(
```
mdbenito, line 419: This is the new pattern: create an empty result and `update` it

src/pydvl/value/shapley/montecarlo.py
```diff
@@ -216,117 +362,81 @@ def permutation_montecarlo_shapley(
...
+        algorithm="combinatorial_montecarlo_shapley", indices=indices
+    )
 
-    values = np.zeros(n)
-    variances = np.zeros(n)
-    counts = np.zeros(n, dtype=int)
-    pbar = maybe_progress(indices, progress, position=job_id)
-    for idx in pbar:
+    repeat_indices = takewhile(lambda _: not done(result), cycle(indices))
+    pbar = tqdm(disable=not progress, position=job_id, total=100, unit="%")
```
mdbenito, line 424: same uglyness, need something better

src/pydvl/value/stopping.py
```diff
@@ -0,0 +1,406 @@
...
+
+    def __invert__(self) -> "StoppingCriterion":
+        return make_criterion(
+            fun=lambda result: ~self._check(result),
+            completion=lambda: 1 - self.completion(),
+            name=f"Composite StoppingCriterion: NOT {self.name}",
+        )(modify_result=self.modify_result)
+
+
+def make_criterion(
```
mdbenito, line 131: I think this is no longer needed, but I am leaving it here for a while, to see if it's useful. I used to do everything with a variation of this function, returning functions wrapped to have binary ops, and the code was much leaner, but I had to start doing weird things to keep state around and decided for objects instead

tests/value/shapley/test_montecarlo.py
```diff
@@ -28,36 +26,41 @@
...
+        (8, ShapleyMode.CombinatorialMontecarlo, 0.2, {"done": MaxUpdates(2**10)}),
+        (
+            12,
+            ShapleyMode.TruncatedMontecarlo,
+            0.1,
+            dict(
+                coordinator_update_period=1,
+                worker_update_period=0.5,
+                done=MaxUpdates(500),
+                truncation=NoTruncation(),
```
mdbenito, line 42: Testing TMCS with truncations is tricky. Lower values lose a lot of accuracy, even if rank is preserved. I think we might need more tests for that. I still need to add tests for the truncation methods themselves 😓 . I have done some eyeball-testing with pdb in `ray debug`, but suppose that's not enough 😄 .

tests/value/shapley/test_naive.py
```diff
@@ -116,10 +116,11 @@ def test_linear_with_outlier(
         scoring=scorer,
         cache_options=MemcachedConfig(client_config=memcache_client_config),
     )
-    shapley_values = permutation_exact_shapley(linear_utility, progress=False)
-    check_total_value(linear_utility, shapley_values, atol=total_atol)
+    values = permutation_exact_shapley(linear_utility, progress=False)
+    values.sort()
```
mdbenito, line 120: values must be explicitly sorted now.

mdbenito (commented):
@AnesBenmerzoug, @Xuzzo  I left some comments to make the review a bit easier, but we can have a meeting if you prefer

src/pydvl/utils/parallel/map_reduce.py
AnesBenmerzoug: It did work for me. Which python version were you using?

The `Iterable` class from collections.abc is not the same as the one from typing.

src/pydvl/value/result.py
AnesBenmerzoug: This class is becoming really big... Can't most of this be done using dataframes?

mdbenito: Yes it can, and it would be much easier and robust and flexible, see my comment. But I don't have the energy to rewrite everything _again_... If it works, let's not touch it, until we need some other feature and someone feels like redoing all of it.

src/pydvl/utils/parallel/map_reduce.py
mdbenito: This breaks as soon as the remote function has string or some container which is not an NDArray as argument, and leads to an infinite recursion. I've hit the problem multiple times and IIRC it breaks with all 3 versions. I know it does with 3.8 on my laptop and 3.10 for sure.

AnesBenmerzoug (approved):

PR #257, opened 2023-02-05
Parallel solution of several LPs in MCLC

### LINKED ISSUE #254: epsilon parameter in MCLC unnecessary
Adding a slack $\epsilon > 0$ is equivalent to requiring $e \ge \epsilon$. But this forces a worse solution by reducing the set of $e$'s over which one minimises. If $e \ge \epsilon$ is a solution it will be found without adding that constraint. If $e \lt \epsilon$ is a solution, then adding the slack variable precludes finding it

mdbenito:
### Description

This PR splits preparation and solution of the linear and quadratic problems in exact and MC least core. This allows batching of the solution of linear problems when running multiple experiments. It also closes #254

### Changes

- Adds `lc_solve_problem`, `lc_solve_problems`, `mclc_prepare_problem`
- Closes #254 

### Checklist

- [x] Updated Changelog
- [x] ~If notebooks were added/changed, added boilerplate cells are tagged with `"nbsphinx":"hidden"~


src/pydvl/value/least_core/__init__.py
```diff
@@ -4,6 +4,70 @@
...
+:func:`~pydvl.value.least_core.montecarlo.mclc_prepare_problem` to prepare a
+list of problems to solve, then solve them in parallel with
+:func:`~pydvl.value.least_core.common.lc_solve_problems`.
+
 """
+from enum import Enum
+from typing import Optional
+
+from ...utils import Utility
+from .. import ValuationResult
```
AnesBenmerzoug, line 22: I find this one much more readable:

```suggestion
from pydvl.value.results import ValuationResult
```

src/pydvl/value/least_core/common.py
```diff
@@ -9,18 +9,127 @@
 __all__ = [
     "_solve_least_core_linear_program",
     "_solve_egalitarian_least_core_quadratic_program",
+    "lc_solve_problem",
+    "lc_solve_problems",
+    "LeastCoreProblem",
 ]
 
+from pydvl.utils import MapReduceJob, Status, Utility
```
AnesBenmerzoug, line 17: These imports should be put before the all dunder variable.

AnesBenmerzoug:
I think it would be good to add tests for the additional functions you defined.

AnesBenmerzoug (approved):

mdbenito:
@AnesBenmerzoug I think this is good to go now 🤞🏽 after adding some tests as suggested (there were lots of bugs 😅) . We should merge once they run and release 0.4.1 while we wait for all the other stuff to be reviewed

AnesBenmerzoug:
@mdbenito looking at the amount of changes in this and other PRs, I think it would be better to directly release 0.5.0

mdbenito:
> @mdbenito looking at the amount of changes in this and other PRs, I think it would be better to directly release 0.5.0

I suggested it because I don't know how long the rest is going to take, given that everybody is busy. Also, we still have the problem with mismatching docs between dev and what's published...

PR #264, opened 2023-02-18
Fixes for GTShapley and scorers

### LINKED ISSUE #263: Owen does not return proper status
should be `Converged`

mdbenito:
### Warning

This PR must be merged after #265 

### Description

This PR revolves around fixes for Group Testing Shapley. PR #267 will add some missing tests (e.g. `Scorer`)

### Changes

- Introduces `utils/score.py` and the `Scorer` type to include information about range and default values when using `compose_score`. This is need e.g. for Group Testing, where an accurate range is necessary.
- Adds a missing argument to GTS
- Switches the default solver to cvxpy's SCS because scipy linprog's `highs` would fail to find solutions in the presence of outlier data.
- Removes the dedicated file `test_gt.py` and adds more tests using group testing.
- Closes #263.

### Checklist

- [x] Updated Changelog


mdbenito:
PR #267 also fixes the MC test with analytic solution for Group Testing which is failing here. I can merge that branch into this one, or we can merge one right after the other since it's based off the top of this one.

docs/30-data-valuation.rst
```diff
@@ -369,14 +397,15 @@ $$
...
 
 Usage follows the same pattern as every other Shapley method, but with the
-addition of an ``eps`` parameter required for the solution of the CSP. It should
-be the same value used to compute the minimum number of samples required. This
-can be done with :func:`~pydvl.value.shapley.gt.num_samples_eps_delta`, but note
-that the number returned will be huge! In practice, fewer samples can be enough,
-but the actual number will strongly depend on the utility, in particular its
-variance.
+addition of an ``epsilon`` parameter required for the solution of the CSP. It
+shouldbe the same value used to compute the minimum number of samples required.
```
mdbenito, line 404: Damn... typo. Will fix somewhere else

AnesBenmerzoug: You don't want to fix it here?

```suggestion
should be the same value used to compute the minimum number of samples required.
```

src/pydvl/value/shapley/gt.py
```diff
@@ -227,15 +230,15 @@ def group_testing_shapley(
     n = len(u.data.indices)
     const = _constants(
         n=n,
-        epsilon=eps,
-        delta=0.05,
+        epsilon=epsilon,
+        delta=delta,
         utility_range=u.score_range.max() - u.score_range.min(),
```
AnesBenmerzoug, line 235: Should we raise an error if group testing is used but no range is passed?

src/pydvl/utils/score.py
```diff
@@ -0,0 +1,126 @@
...
+    def __init__(
+        self,
+        scoring: Union[str, ScorerCallable],
+        default: float = np.nan,
+        range: Tuple = (-np.inf, np.inf),
+        name: Optional[str] = None,
+    ):
+        self._scorer = get_scorer(scoring)
+        self.default = default
+        # TODO: auto-fill from known scorers ?
```
AnesBenmerzoug, line 65: Should we just hard-code it?

We can get the names of the scorers from [here](https://github.com/scikit-learn/scikit-learn/blob/8c9c1f27b/sklearn/metrics/_scorer.py#L842).

I will create an issue for this but since it's only used for group testing it doesn't really have a high priority.

src/pydvl/value/shapley/gt.py
```diff
@@ -277,56 +281,66 @@ def reducer(
 
     ###########################################################################
     # Solution of the constraint problem with scipy
+    # DISABLED: highs fails to find solutions that SCS finds
```
AnesBenmerzoug, line 284: I think we should remove this since the cvxpy approach is more readable and more stable.

src/pydvl/value/shapley/gt.py
```diff
@@ -277,56 +281,66 @@ def reducer(
...
-        b_ub=b_ub,
-        A_eq=np.ones((1, n)),
-        b_eq=total_utility,
-        bounds=bounds,
-        method="highs",
-        options={},
+    ###########################################################################
+    # Solution of the constraint problem with cvxpy
+
+    import cvxpy as cp
```
AnesBenmerzoug, line 304: Is there a reason for this import being here?

src/pydvl/value/shapley/gt.py
```diff
@@ -277,56 +281,66 @@ def reducer(
...
+
+    v = cp.Variable(n)
+    constraints = [cp.sum(v) == total_utility]
+    for i in range(n):
+        for j in range(i + 1, n):
+            constraints.append(v[i] - v[j] <= epsilon + C[i, j])
+            constraints.append(v[j] - v[i] <= epsilon - C[i, j])
+
+    problem = cp.Problem(cp.Minimize(0), constraints)
+    problem.solve(solver=cp.SCS)
```
AnesBenmerzoug, line 314: We should allow the user to change just like in the least core methods.

AnesBenmerzoug (approved):
Other than the one comment and one question that I left, this looks good to me and could be merged as is.

src/pydvl/value/shapley/gt.py
mdbenito: Or maybe we can just not try to compute a value for `max_iterations` and leave it to the user. I'll warn instead

src/pydvl/value/shapley/gt.py
mdbenito: hmm... but what if we end up deciding to go for a linear system again... I will leave a note of when this was removed, so as to be able to find it in the history.

src/pydvl/value/shapley/gt.py
mdbenito: yeah, this was not a dependency before and I was double checking that I was creating the matrices for the linear problem correctly by solving with a different method. I will move it up.

src/pydvl/value/shapley/gt.py
mdbenito: On it

PR #278, opened 2023-02-24
IF refactoring and further testing

### LINKED ISSUE #130: Investigate flakiness of influence values

### LINKED ISSUE #137: Batched conjugate gradient is brittle
I have found several times that the batched cg does not converge while the scipy cg converges. I believe there is something strange going on with the step size. However, even in the cases in which it converges, it is slower than the non-batched one, so for now we can use the scipy implementation and leave this for when we are benchmarking runtime.

### LINKED ISSUE #168: Review notebooks on influence functions
That.

### LINKED ISSUE #169: Unit tests for influence functions
Ensure coverage

### LINKED ISSUE #203: Label all IF code and documentation as experimental / unreliable
Before 1st official release

### LINKED ISSUE #222: Inspect use of Pytorch lightning to replace TorchWrappers
We have implemented quite a few pytorch convenience wrappers to simplify interface in notebooks. However, there are several library that already to that. Investigate whether it makes sense (it probably does) to replace our wrappers with the pytorch-lightning ones.

### LINKED ISSUE #306: Cleanup influence tests
Some tests for influence functions use `itertools.product` on a bunch of static vars to create tests matrices. Instead this should be done using stacked calls to `@pytest.mark.parametrize`, without any vars

Xuzzo:
### Description

This PR closes #137 and closes #169 and closes #203 and closes #130 and closes #168 and closes #306
It also closes #222: wrappers have been removed and only one fitting method is used, now outside src (put in notebook_support).

### Changes

- Several refactorings of influence functions code
- included if tests with several NN architectures and more detailed ones using linear models
- removed useless framework wrappers and boilerplate from source code
- Re-runs notebooks with new (slightly) modified interface

### Checklist

- [x] Updated Changelog


src/pydvl/influence/types.py
```diff
@@ -1,48 +1,53 @@
...
+try:
+    import torch
+    import torch.nn as nn
+
+    _TORCH_INSTALLED = True
+except ImportError:
+    _TORCH_INSTALLED = False
+
+if TYPE_CHECKING:
+    from numpy.typing import NDArray
```
AnesBenmerzoug, line 15: We decided in a previous discussion to import this directly instead of only during type checking because it doesn't add anything to do it this way.

notebooks/notebook_support.py
```diff
@@ -127,10 +274,25 @@ def load(self) -> Tuple["NDArray[np.float_]", "NDArray[np.float_]"]:
...
+    ds_label_to_model_label = {
+        ds_label: idx for idx, ds_label in enumerate(labels.values())
+    }
+    device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+    x_nn = torch.stack(x.tolist()).to(device)
+    y_nn = torch.tensor([ds_label_to_model_label[yi] for yi in y], device=device)
+    return x_nn, y_nn
+
+
 def plot_dataset(
```
mdbenito, line 296: This seems like a very unfortunate name. Does the function plot _any_ dataset?

notebooks/notebook_support.py
```diff
@@ -127,10 +274,25 @@ def load(self) -> Tuple["NDArray[np.float_]", "NDArray[np.float_]"]:
...
     train_ds: Tuple["NDArray[np.float_]", "NDArray[np.int_]"],
```
mdbenito, line 297: Can't we use our `Dataset`? (asking)

mdbenito (commented):
I've added a few comments in two files but I'm calling it a day. Will continue soon

src/pydvl/influence/types.py
```diff
@@ -1,48 +1,53 @@
...
+
+
+__all__ = ["TwiceDifferentiable", "TensorType"]
+
+TensorType = Union["NDArray", "torch.TensorType"]
```
AnesBenmerzoug, line 15: I didn't know about `torch.TensorType`. It seems it was added in [this commit](https://github.com/pytorch/pytorch/commits/9f3167e) and should be part of Pytorch 1.10.

We should either provide an alternative if it doesn't exist, limit pytorch version to those after 1.10 or use `torch.Tensor` instead.

tests/influence/test_torch_differentiable.py
```diff
@@ -10,31 +10,29 @@
...
 
 try:
+    import torch
     import torch.nn.functional as F
+    from torch import nn
 
     from pydvl.influence.frameworks import TorchTwiceDifferentiable
-    from pydvl.influence.model_wrappers import TorchLinearRegression
 except ImportError:
     pass
```
AnesBenmerzoug, line 27: We can replace this try except block with:

```suggestion
torch = pytest.importorskip("torch")

import torch.nn.functional as F
from pydvl.influence.frameworks import TorchTwiceDifferentiable
from pydvl.influence.model_wrappers import TorchLinearRegression
```

This is will skip this test file if pytorch cannot be imported.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 0.6.0 - Refactorings
```
mdbenito, line 3: Because one can't be sure of the release number beforehand, the convention is to title the section "unreleased":
```suggestion
## Unreleased
```

tests/influence/conftest.py
```diff
@@ -1,11 +1,18 @@
...
 from pydvl.utils import Dataset, random_matrix_with_condition_number
 
+try:
+    import torch
+    from torch import nn
+except ImportError:
+    pass
+
 if TYPE_CHECKING:
     from numpy.typing import NDArray
```
mdbenito, line 11:
```suggestion
from numpy.typing import NDArray
```

tests/influence/conftest.py
```diff
@@ -75,7 +52,152 @@ def linear_model(problem_dimension: Tuple[int, int], condition_number: float):
     return A, b
 
 
-def create_mock_dataset(
+def linear_derivative_analytical(
+    linear_model: Tuple["NDArray", "NDArray"], x: "NDArray", y: "NDArray"
+) -> "NDArray":
```
mdbenito, line 51:
```suggestion
    linear_model: Tuple[NDArray[np.float_], NDArray[np.float_]], x: NDArray[np.float_], y: NDArray[np.float_]
) -> NDArray[np.float_]:
```

tests/influence/conftest.py
```diff
@@ -75,7 +52,152 @@ def linear_model(problem_dimension: Tuple[int, int], condition_number: float):
...
+    """
+    :param linear_model: A tuple of np.ndarray' of shape [NxM] and [N] representing A and b respectively.
+    :param x: A np.ndarray of shape [BxM].
+    :param y: A np.nparray of shape [BxN].
+    :returns: A np.ndarray of shape [Bx((N+1)*M)], where each row vector is [d_theta L(x, y), d_b L(x, y)]
```
mdbenito, line 56: It is unnecessary to repeat the type in the param docstring, since it's in the function's signature.
```suggestion
    :param linear_model: A tuple of matrices of shape [NxM] and [N] representing A and b respectively.
    :param x: A matrix of shape [BxM].
    :param y: A matrix of shape [BxN].
    :returns: A matrix of shape [Bx((N+1)*M)], where each row vector is [d_theta L(x, y), d_b L(x, y)]
```
Also, it's better to use LaTeX (🙈) for the math. Sphinx renders it and it can be less ambiguous (what is `d_theta L(x,y)`? Better to have the definition of $L(x,y)$ in the docstring to make the dependence on $\theta, b$ clear)

tests/influence/conftest.py
```diff
@@ -1,11 +1,18 @@
...
 from sklearn.preprocessing import MinMaxScaler
 
+from pydvl.influence.general import InfluenceType
 from pydvl.utils import Dataset, random_matrix_with_condition_number
 
+try:
+    import torch
+    from torch import nn
+except ImportError:
+    pass
```
mdbenito: See Anes' comment below on `importorskip`

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -49,7 +51,12 @@ def __init__(
         """
         if not _TORCH_INSTALLED:
             raise RuntimeWarning("This function requires PyTorch.")
```
mdbenito, line 53: This feels a bit awkward, especially if it has to be done in each function using torch. Instead, one could try the imports in the `__init__.py` and import the right submodule based on that. E.g.
```python

try:
    import torch
    _TORCH_INSTALLED = True
except ImportError:
    _TORCH_INSTALLED = False
try:
    import jax
    _JAX_INSTALLED = True
except ImportError:
    _JAX_INSTALLED = False

if _TORCH_INSTALLED and FRAMEWORK == "torch":
    from frameworks.torch_differentiable import TorchTwiceDifferentiable as TwiceDifferentiable
    # etc. (or something like that)
elif _JAX_INSTALLED and FRAMEWORK == "jax":
     from frameworks.jax_differentiable import JaxTwiceDifferentiable as TwiceDifferentiable
     #...
```
This is however not very nice, because one must set the `FRAMEWORK` beforehand, maybe with an env variable or before importing the parent module, so that sucks.

Alternatively we could have submodules like:
```
influence/
|__torch.py
|__jax.py
|__tensorflow.py
|__ _common.py
```
and users would do
```python
import influence.torch as inf
inf.compute_influences(torch tensors and stuff)
```
Or 
```python
import influence.jax as inf
inf.compute_influences(jax data)
```
This will probably mean a lot of code duplication, though. And across different files. Using something like eagerpy might be tempting but those things just tend to die once the phd student defends or whatever (like eagerpy seems to have done).

Alternatively we could write some kind of `@dispatch` alternative, based on imports (or dispatch itself, based on the tensor types?) and keep all functions in the same file, so that at least they are bunched together...

Dunno, just spitballing here.. Ideas @AnesBenmerzoug @Xuzzo ?

mdbenito (commented):
Added a few more comments and some discussion on how to avoid the checks for modules inside the functions, which might pave the way towards #282

notebooks/notebook_support.py
```diff
@@ -28,8 +43,143 @@
 
 MODEL_PATH = Path().resolve().parent / "data" / "models"
 
+logger = logging.getLogger(__name__)
 
-def new_resnet_model(output_size: int) -> TorchModel:
+
+class TorchLogisticRegression(nn.Module):
```
mdbenito, line 49: Won't this fail if torch is not installed? I think all torch-dependent methods need to be factored out into their own module, which is only imported when torch is available.

notebooks/notebook_support.py
Xuzzo: The Dataset class is not optimal for torch datasets at the moment. Main issues:
- it does not have validation set
- it runs checks with sklearn which would need to convert torch tensors, with memory overhead
- Apart for wrapping train and test sets, it does not add anything more since its methods are not used in influences (e.g. subsampling of indices)
- Not compatible with torch DataLoaders

src/pydvl/influence/general.py
```diff
@@ -77,17 +61,24 @@ def calculate_influence_factors(
...
-    influence_factors: "NDArray",
+    x: TensorType,
+    y: TensorType,
+    influence_factors: NDArray,
     progress: bool = False,
-) -> "NDArray":
+) -> NDArray:
     """
     Calculates the influence from the influence factors and the scores of the training points.
     Uses the upweighting method, as described in section 2.1 of https://arxiv.org/pdf/1703.04730.pdf
```
mdbenito, line 84: See my comment before: use a bib. reference, and document what's happening

src/pydvl/influence/general.py
```diff
@@ -105,11 +96,11 @@ def _calculate_influences_up(
 
 def _calculate_influences_pert(
     model: TwiceDifferentiable,
-    x: "NDArray",
-    y: "NDArray",
-    influence_factors: "NDArray",
+    x: TensorType,
```
mdbenito, line 99: See my comment below about union types and how they fail to catch mixed argument types. If you want to accept both values, you can use `@overload` for the benefit of the type checker. Or a generic function with a TypeVar constrained to the two types (but I'm not sure if that would catch the mixed type problem).

Here's an example for the first:

```python
@overload
def _calculate_influences_pert(model: NumpyTwiceDifferentiable, x: NDArray[np.float_], blahblah) -> NDArray[np.float_]:
    ...
@overload
def _calculate_influences_pert(model: TorchTwiceDifferentiable, x: torch.Tensor, blahblah) -> torch.Tensor:
    ...

def _calculate_influences_pert(model: Union[], x: Union[]...) -> Union[...]:
    # schtuff

**Review by mdbenito — COMMENTED** [review]
Added a few more comments about types and so on.

File: src/pydvl/influence/frameworks/torch_differentiable.py (hunk 11 above)
Comment (Xuzzo) [reply src/pydvl/influence/frameworks/torch_differentiable.py]: I went for the env variable. Let me know if it looks good enough

File: src/pydvl/influence/inversion_methods.py [hunk 15]
```diff
@@ -0,0 +1,82 @@
...
+    Different inversion methods types.
+    """
+
+    Direct = "direct"
+    Cg = "cg"
+
+
+def invert_matrix(
+    inversion_method: InversionMethod,
+    mvp: Callable[[NDArray], NDArray],
```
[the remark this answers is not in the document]
Xuzzo: Should be solved.

src/pydvl/influence/general.py
Xuzzo: should be solved with the new changes.

notebooks/notebook_support.py
```diff
@@ -1,147 +1,33 @@
...
-from torch.optim import Adam
-from torchvision.models import ResNet18_Weights, resnet18
 
-from pydvl.influence.model_wrappers import TorchModel
 from pydvl.utils import Dataset
 
-try:
-    import torch
+logger = logging.getLogger(__name__)
+Losses = namedtuple("Losses", "training val")
```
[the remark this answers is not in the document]
Xuzzo: I have kept the namedtuple in the end. I think it does not hurt too much...

src/pydvl/influence/general.py
```diff
@@ -39,26 +33,16 @@ class InfluenceType(str, Enum):
...
+    y: TensorType,
+    x_test: TensorType,
+    y_test: TensorType,
+    inversion_method: InversionMethod,
     lam: float = 0,
     progress: bool = False,
-) -> "NDArray":
+) -> NDArray:
     """
     Calculates the influence factors. For more info, see https://arxiv.org/pdf/1703.04730.pdf, paragraph 3.
```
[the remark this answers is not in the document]
Xuzzo: should be done

Xuzzo:
I removed the env variable, as we discussed.

src/pydvl/influence/frameworks/torch_differentiable.py
Xuzzo: Better solution proposed here #297

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -1,29 +1,28 @@
...
     from torch import autograd
     from torch.autograd import Variable
 
-    _TORCH_INSTALLED = True
 except ImportError:
-    _TORCH_INSTALLED = False
-
-if TYPE_CHECKING:
-    from numpy.typing import NDArray
+    pass
```
mdbenito, line 20: 🤔 How can this work? If the code below relies on torch symbols imported here, if the import fails, the interpreter will choke. Have you tested in an environment without torch? If you want to make the stuff here available from `__init__.py`, the checking must be done there. Maybe try/except importing this module, and if you succeed, append the necessary symbols to `__all__` of that module.

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -33,23 +32,26 @@ def flatten_gradient(grad):
...
         self,
-        model: "nn.Module",
-        loss: Callable[["torch.Tensor", "torch.Tensor"], "torch.Tensor"],
+        model: nn.Module,
+        loss: Callable[[torch.Tensor, torch.Tensor], torch.Tensor],
     ):
         """
-        :param model: A torch.nn.Module representing a (differentiable) function f(x).
-        :param loss: Loss function L(f(x), y) maps a prediction and a target to a single value.
+        :param model: A torch.nn.Module representing a (differentiable) function.
```
mdbenito, line 46: As mentioned before, the type is redundant
```suggestion
        :param model: A (differentiable) function.
```

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -63,18 +65,19 @@ def num_params(self) -> int:
...
-    ) -> "NDArray":
+    ) -> NDArray:
         """
-        Calculates gradient of model parameters wrt each x[i] and y[i] and then
+        Calculates gradient of model parameters wrt each $x[i]$ and $y[i]$ and then
         returns a array of size [N, P] with N number of points (length of x and y) and P
         number of parameters of the model.
-        :param x: A np.ndarray [NxD] representing the features x_i.
-        :param y: A np.ndarray [NxK] representing the predicted target values y_i.
+        :param x: A np.ndarray [NxD] representing the features $x_i$.
```
mdbenito, line 76: A docstring type is redundant because it is already available in the signature, and in this case, it contradicts it

src/pydvl/influence/frameworks/twice_differentiable.py
```diff
@@ -0,0 +1,54 @@
...
+        """
+        Calculate the gradient of the model wrt each input x and labels y.
+        The output is therefore of size [Nxp], with N the amout of points (the
+        length of x and y) and P the number of parameters.
+        """
+        pass
+
+    def grad(self, x: TensorType, y: TensorType) -> Tuple[NDArray, TensorType]:
+        """
+        It calculates the gradient of model parameters with respect to input x
```
mdbenito, line 36: A pronoun is strange here. The subject is just beside the verb. ("grad(...): calculates"). There are more examples.

```suggestion
        Calculates the gradient of model parameters with respect to input x
```

src/pydvl/influence/frameworks/twice_differentiable.py
```diff
@@ -0,0 +1,54 @@
...
+        progress: bool = False,
+    ) -> NDArray:
+        """
+        Calculate the gradient of the model wrt each input x and labels y.
+        The output is therefore of size [Nxp], with N the amout of points (the
+        length of x and y) and P the number of parameters.
+        """
+        pass
+
+    def grad(self, x: TensorType, y: TensorType) -> Tuple[NDArray, TensorType]:
```
mdbenito, line 34: The return type is not documented. And why an NDArray suddenly?

src/pydvl/influence/general.py
```diff
@@ -39,88 +22,107 @@ class InfluenceType(str, Enum):
...
 
 
 def _calculate_influences_up(
     model: TwiceDifferentiable,
-    x: "NDArray",
-    y: "NDArray",
-    influence_factors: "NDArray",
+    x: TensorType,
+    y: TensorType,
+    influence_factors: NDArray,
```
mdbenito, line 76: Why not a TensorType as the rest? It would make more sense IMO to remain within the framework used.

src/pydvl/influence/general.py
```diff
@@ -39,88 +22,107 @@ class InfluenceType(str, Enum):
...
+        inversion_method,
+        hvp,
+        (n_params, n_params),
+        test_grads,
+        progress,
+    )
 
 
 def _calculate_influences_up(
     model: TwiceDifferentiable,
```
mdbenito, line 73: Shouldn't this be
```suggestion
    model: TwiceDifferentiable[TensorType, ModelType],
```

src/pydvl/influence/frameworks/twice_differentiable.py
```diff
@@ -0,0 +1,54 @@
...
+        """
+        pass
+
+    def mvp(
+        self,
+        grad_xy: TensorType,
+        v: TensorType,
+        progress: bool = False,
+        backprop_on: Optional[TensorType] = None,
+    ) -> NDArray:
```
mdbenito, line 47: Same as before: mixing types is strange (and possibly inefficient if moving memory from gpu to cpu)

mdbenito (commented):
Added more comments. A general guideline: try remaining within the same data type when working, except in explicit conversion functions. TensorType in: TensorType out. This: makes working completely within the same device possible (maybe), reduces copying (conversions involve lots of copying), but also makes the interfaces more predictable. And more generally, consistency is key for predictability and maintainability, also in variable names (even the order in which nouns and verbs appear), documentation, parameter grouping by "category" in function signatures, etc.

notebooks/support/types.py
```diff
@@ -0,0 +1,3 @@
+from collections import namedtuple
+
+Losses = namedtuple("Losses", "training val")
```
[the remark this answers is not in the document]
mdbenito: Oh I just saw my other comment. Actually, the type is redefined, which is not good. If Losses will depend on the framework (it probably will, right?), then it must be typed as well with `TensorType`

src/pydvl/influence/general.py
```diff
@@ -2,32 +2,15 @@
 Contains parallelized influence calculation functions for general models.
```
mdbenito, line 2: As opposed to functions, here a subject _is_ needed 😅 

Also, this module should be called "common" for consistency with the rest of the library

src/pydvl/influence/general.py
```diff
@@ -2,32 +2,15 @@
...
-    _TORCH_INSTALLED = True
-except ImportError:
-    _TORCH_INSTALLED = False
-
-if TYPE_CHECKING:
-    from numpy.typing import NDArray
-
-
-__all__ = ["compute_influences", "InfluenceType", "InversionMethod"]
+__all__ = ["compute_influences", "InfluenceType", "calculate_influence_factors"]
```
mdbenito, line 13: Again, consistency: either "compute" or "calculate", pick one everywhere! Value uses "compute", so I'd stick with that

src/pydvl/influence/inversion_methods.py
```diff
@@ -0,0 +1,84 @@
...
+        returns the result of $Ax$.
+    :param mvp_dimensions: dimensions of matrix $A$
+    :param b:
+    :param progress: If True, display progress bars.
+
+    :return: An array that solves the inverse problem,
+        i.e. it returns $x$ such that $Ax = b$
+    """
+    if inversion_method == InversionMethod.Direct:
+        return (np.linalg.inv(mvp(np.eye(mvp_dimensions[1]))) @ b.T).T  # type: ignore
```
mdbenito, line 57: Maybe we discussed this already, but this should use `torch.linalg.inv` if the framework is torch, etc. Otherwise you lose the GPU advantage, copy memory around like crazy (very expensive here) and so on.

src/pydvl/influence/inversion_methods.py
```diff
@@ -0,0 +1,84 @@
...
+    Cg = "cg"
+
+
+def invert_matrix(
+    inversion_method: InversionMethod,
+    mvp: Callable[[TensorType], NDArray[np.float_]],
+    mvp_dimensions: Tuple[int, int],
+    b: NDArray[np.float_],
+    progress: bool = False,
+) -> NDArray[np.float_]:
```
mdbenito, line 38: This should return TensorType as well, thus making `compute_influences` also consistent with types

tests/influence/test_torch_differentiable.py
```diff
@@ -10,31 +10,27 @@
...
+from torch import nn
 
-    from pydvl.influence.frameworks import TorchTwiceDifferentiable
-    from pydvl.influence.model_wrappers import TorchLinearRegression
-except ImportError:
-    pass
+from pydvl.influence.frameworks.torch_differentiable import TorchTwiceDifferentiable
 
 
 class ModelTestSettings:
```
mdbenito, line 28: We don't use classes to store parameters in the other test suites. Reasons:
1. Yuk!
2. `pytest.mark.parametrize` allows tuning for each test and is more explicit
3. Classes are data types, not data collections. If this were to be done (it won't), `ModelTestSettings` ought to be a dataclass or NamedTuple and an _instance_ would contain the values. Using class values for configuration is a pointless use of a class and unexpected for anyone seeing it.

AnesBenmerzoug: This was already here before so I suggest we create an issue and do this in a separate PR otherwise this won't be merged any time soon.

docs/pydvl.bib
```diff
@@ -50,6 +50,22 @@ @article{jia_efficient_2019a
...
+  archiveprefix = {arXiv},
+  langid = {english},
+  file = {/Users/fabio/Zotero/storage/HW6BK957/D9MZ9MUI.pdf;/Users/fabio/Zotero/storage/J4YQYLEN/Koh and Liang - 2017 - Understanding Black-box Predictions via Influence .pdf;/Users/fabio/Zotero/storage/MHFVLFHJ/Koh and Liang - 2017 - Understanding Black-box Predictions via Influence .pdf}
```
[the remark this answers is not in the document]
Xuzzo: should be in zotero too.

tests/influence/test_torch_differentiable.py
Xuzzo: I have removed the class. Not sure if here you also meant to write each input specifically instead of through itertools

src/pydvl/influence/inversion_methods.py
Xuzzo: types should be fine now

src/pydvl/influence/inversion_methods.py
Xuzzo: using torch.linalg.solve now, which is recommended over linalg.inv (https://pytorch.org/docs/stable/generated/torch.linalg.inv.html)

src/pydvl/influence/frameworks/twice_differentiable.py
Xuzzo: should be fine now

notebooks/support/common.py
```diff
@@ -1,147 +1,33 @@
...
-from torch.optim import Adam
-from torchvision.models import ResNet18_Weights, resnet18
 
-from pydvl.influence.model_wrappers import TorchModel
 from pydvl.utils import Dataset
 
-try:
-    import torch
+logger = logging.getLogger(__name__)
+Losses = namedtuple("Losses", "training val")
```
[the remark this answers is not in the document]
Xuzzo: should be solved

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -33,93 +26,164 @@ def flatten_gradient(grad):
     return torch.cat([el.reshape(-1) for el in grad])
 
 
-class TorchTwiceDifferentiable(TwiceDifferentiable):
+def solve_linear(matrix: torch.Tensor, b: torch.Tensor):
+    """Computes the solution of a square system of linear equations"""
+    return torch.linalg.solve(matrix, b)
+
+
+def as_tensor(a: Any, ensure=True, **kwargs):
```
mdbenito, line 34: `ensure` seems to be a misnomer, since the conversion is always ensured, but this flag decides whether a warning is issued. Maybe `warn`? (And docstring missing)

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -33,93 +26,164 @@ def flatten_gradient(grad):
...
+            flatten_gradient(autograd.grad(z[i], backprop_on, retain_graph=True))
+        )
+    mvp = torch.stack([grad.contiguous().view(-1) for grad in mvp])
+    return mvp.detach()  # type: ignore
+
+
+class TorchTwiceDifferentiable(TwiceDifferentiable[torch.Tensor, nn.Module]):
     """
-    Calculates second-derivative matrix vector products (Mvp) of a pytorch torch.nn.Module
+    Calculates second-derivative of a model wrt. a given loss
```
mdbenito, line 98: This generic description belongs in the Base class, I'd say. Also, I would write something like
"Wraps a differentiable model and loss and provides methods to compute the second derivative of the loss wrt. the model parameters" (if this is what is happening. At any rate the derivative of the model wrt a loss doesn't really make sense I think)

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -33,93 +26,164 @@ def flatten_gradient(grad):
...
-        model: "nn.Module",
-        loss: Callable[["torch.Tensor", "torch.Tensor"], "torch.Tensor"],
+        model: nn.Module,
+        loss: Callable[[torch.Tensor, torch.Tensor], torch.Tensor],
     ):
         """
-        :param model: A torch.nn.Module representing a (differentiable) function f(x).
-        :param loss: Loss function L(f(x), y) maps a prediction and a target to a single value.
+        :param model: A (differentiable) function.
+        :param loss: Loss function $L(f(x), y)$ maps a prediction and a target to a single value.
```
mdbenito, line 108:
```suggestion
        :param loss: A differentiable scalar loss $L(\hat{y}, y)$, mapping a prediction and a target to a real value.
```

src/pydvl/influence/frameworks/twice_differentiable.py
```diff
@@ -0,0 +1,52 @@
...
+        """Returns all the model parameters that require differentiation"""
+        pass
+
+    def split_grad(
+        self,
+        x: TensorType,
+        y: TensorType,
+        progress: bool = False,
+    ) -> TensorType:
+        """
```
mdbenito, line 30: Parameters undocumented

src/pydvl/influence/frameworks/twice_differentiable.py
```diff
@@ -0,0 +1,52 @@
...
+        Calculate the gradient of the model wrt each input x and labels y.
+        The output is therefore of size [Nxp], with N the amount of points (the
+        length of x and y) and P the number of parameters.
+        """
+        pass
+
+    def grad(self, x: TensorType, y: TensorType) -> Tuple[TensorType, TensorType]:
+        """
```
mdbenito, line 38: Parameters undocumented, in particular the return type! (it's unexpected)

src/pydvl/influence/frameworks/twice_differentiable.py
```diff
@@ -0,0 +1,52 @@
...
+        pass
+
+    def hessian(
+        self,
+        x: TensorType,
+        y: TensorType,
+        progress: bool = False,
+    ) -> TensorType:
+        """Calculates the explicit hessian of model parameters given data ($x$
+        and $y$)."""
```
mdbenito, line 51: Maybe you mean this?

```suggestion
        """Calculates the full Hessian of $L(f(x),y)$ with respect to the model parameters
           given data ($x$ and $y$)."""
```

src/pydvl/influence/frameworks/twice_differentiable.py
```diff
@@ -0,0 +1,52 @@
...
+
+TensorType = TypeVar("TensorType", bound=Sequence)
+ModelType = TypeVar("ModelType")
+
+
+class TwiceDifferentiable(ABC, Generic[TensorType, ModelType]):
+    def __init__(
+        self,
+        model: ModelType,
+        loss: Callable[[TensorType, TensorType], TensorType],
```
mdbenito, line 12: The more I look at this, the more I wonder, why does something that computes second derivatives of functions need to know about models and losses? The thing that is differentiated is the composition L(f), evaluated at x,y, wrt. the parameters of f, correct? I would expect this wrapper to wrap around a single function, which would be anything. I would then use some `compose` primitive to compose models, losses, whatever. Maybe we can make an issue for this? Or does it not make sense?

src/pydvl/influence/general.py
```diff
@@ -39,88 +28,111 @@ class InfluenceType(str, Enum):
...
-    :param x_train: A np.ndarray of shape [MxK] containing the features of the input data points.
-    :param y_train: A np.ndarray of shape [MxL] containing the targets of the input data points.
-    :param x_test: A np.ndarray of shape [NxK] containing the features of the test set of data points.
-    :param y_test: A np.ndarray of shape [NxL] containing the targets of the test set of data points.
-    :param inversion_func: function to use to invert the product of hvp (hessian vector product) and the gradient
-        of the loss (s_test in the paper).
+) -> TensorType:
+    r"""
+    It calculates influence factors of a model from its training and test
+    points. Given a test point $z_test = (x_{test}, y_{test})$, a loss
```
mdbenito, line 43:
```suggestion
    Calculates influence factors of a model for training and test
    data. Given a test point $z_test = (x_{test}, y_{test})$, a loss
```

src/pydvl/influence/general.py
```diff
@@ -39,88 +28,111 @@ class InfluenceType(str, Enum):
...
+    $L(z_{test}, \theta)$ ($\theta$ being the parameters of the model) and the
+    Hessian of the model $H_{\theta}$, influence factors are defined as
+    $s_{test} = H_{\theta}^{-1} \grad_{\theta} L(z_{test}, \theta)$. They are
+    used for efficient influence calculation. This method first
```
mdbenito, line 47:
```suggestion
    $$s_{test} = H_{\theta}^{-1} \grad_{\theta} L(z_{test}, \theta).$$
    They are used for efficient influence calculation. This method first
```

tests/influence/test_influences.py
```diff
@@ -4,57 +4,105 @@
...
+    :param x: An array of shape (M, K) containing the features of the
+        training set.
+    :param y: An array of shape (M, L) containing the targets of the
+        training set.
+    :param x_test: An array of shape (N, K) containing the features of the test
+        set.
+    :param y_test: An array of shape (N, L) containing the targets of the test
+        set.
+    :param influence_type: the type of the influence.
+    :param hessian_retularization: regularization value for the hessian
```
mdbenito, line 49: Typo:
```suggestion
    :param hessian_regularization: regularization value for the Hessian
```
But also, the docstring just repeats the parameter name.

tests/influence/test_influences.py
```diff
@@ -4,57 +4,105 @@
...
+INFLUENCE_TRAINING_SET_SIZE: List[int] = [50, 30]
+INFLUENCE_TEST_SET_SIZE: List[int] = [20]
+INFLUENCE_DIMENSIONS: List[Tuple[int, int]] = [
+    (10, 10),
+    (3, 20),
+]
+HESSIAN_REGULARIZATION: List[float] = [0, 1]
 
 
 test_cases = list(
```
mdbenito, line 89: As mentioned in #306, this seems like reinventing `@pytest.mark.parametrize`. Just stack calls to this decorator to get a product.

tests/influence/test_torch_differentiable.py
```diff
@@ -0,0 +1,177 @@
...
+TEST_SET_SIZE: List[int] = [20]
+TRAINING_SET_SIZE: List[int] = [50]
+PROBLEM_DIMENSIONS: List[Tuple[int, int]] = [
+    (2, 2),
+    (5, 10),
+    (10, 5),
+    (10, 10),
+]
+
+test_cases_linear_regression_derivatives = list(
```
mdbenito, line 42: Same comment as above about using `@pytest.mark.parametrize`, and maybe do this in #306

mdbenito (commented):
Almost there! 💪🏽

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -33,93 +26,164 @@ def flatten_gradient(grad):
...
+
+
+def as_tensor(a: Any, ensure=True, **kwargs):
+    """Converts an array into a torch tensor"""
+    if ensure and not isinstance(a, torch.Tensor):
+        logger.warning("Converting tensor to type torch.Tensor.")
+    return torch.as_tensor(a, **kwargs)
+
+
+def stack_tensors(a: Sequence[torch.Tensor], **kwargs):
```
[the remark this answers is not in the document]
Xuzzo: should be done

src/pydvl/influence/frameworks/twice_differentiable.py
mdbenito: See #315

mdbenito:
I think this is ready now, given all the things that we have delayed to new issues and the bigger design changes to come. Will merge after releasing 0.5.1

mdbenito (approved):
Lots of thing still to do, but we need to merge

PR #281, opened 2023-02-25
Fix Monte Carlo Least Core error when n_iterations < len(dataset)

### LINKED ISSUE #251: Monte Carlo Least Core raises an error when n_iterations < len(dataset)
## Description

In the current implementation, `montecarlo_least_core()` raises an error if the number of constraints (`n_iterations`) is less than the number of data points (`len(u.dataset)`) even though the default solver from `cvxpy` can find a solution.

This was done because the first implementation used solvers from `scipy` and those failed for that specific case.

## How to reproduce?

```python
import numpy as np
from pydvl.utils import Dataset, Utility
from pydvl.value.least_core import montecarlo_least_core
from sklearn.linear_model import LogisticRegression

X, y = np.arange(20).reshape((10, 2)), np.random.uniform(low=0.0, high=1.0, size=10)
dataset = Dataset.from_arrays(X, y,  train_size=0.5, random_state=16)
utility = Utility(LogisticRegression(), dataset)
values = montecarlo_least_core(u=utility, n_iterations=3)
```

### Expected output

I expected the values to be computed and returned.

### Actual Output

```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/anesbenmerzoug/workdir/repositories/transferlab/pyDVL/.tox/base/lib/python3.8/site-packages/pydvl/value/least_core/montecarlo.py", line 121, in montecarlo_least_core
    raise ValueError(
ValueError: Number of iterations should be greater than the size of the dataset
```

AnesBenmerzoug:
### Description

This PR closes #251

### Changes

- Prints a warning instead of raising an error when `n_iterations` < len(dataset) in Monte Carlo Least Core.
- Moved `montecarlo_least_core` higher in the module for readability.
- Removed test that is no longer needed.

### Checklist

- [X] Updated Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - **Bug fix**: Warn instead of raise an error when `n_iterations`
   is less than the size of the dataset in Monte Carlo Least Core
-  [PR #279](https://github.com/appliedAI-Initiative/pyDVL/pull/279)
+  [PR #281](https://github.com/appliedAI-Initiative/pyDVL/pull/281)
 
 ## 0.5.0 - 💥 Fixes, nicer interfaces and... more breaking changes 😒
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- **Bug fix**: Warn instead of raise an error when `n_iterations`
+- **Bug fix**: Warn instead of raising an error when `n_iterations`
   is less than the size of the dataset in Monte Carlo Least Core
   [PR #281](https://github.com/appliedAI-Initiative/pyDVL/pull/281)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+- **Bug fix**: Warn instead of raise an error when `n_iterations`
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+- **Bug fix**: Warn instead of raise an error when `n_iterations`
```
mdbenito, line 5:
```suggestion
- **Bug fix**: Warn instead of raising an error when `n_iterations`
```

mdbenito (approved):
lgtm, but looking at this, the name `n_iterations` doesn't really make much sense, does it? It's the number of constraints used in the problem. Before, it made sense to use `n_iterations` everywhere for consistency, but now the main MC methods use `StoppingCriterion`. And since LC and GT need a dedicated parameter, which in both cases is a number of constraints or samples to take, maybe we should change the name to be `n_samples` or `n_constraints` in both. (?)

AnesBenmerzoug:
I wanted to tackle the issue of consistency in the argument names in another issue.

I think we have to agree first on what to use and then document it before making changes.

mdbenito:
> I wanted to tackle the issue of consistency in the argument names in another issue.

Created #295

PR #300, opened 2023-03-05
Fix StandardError stopping criterion

### LINKED ISSUE #299: StandardError criterion is bogus
As of 0.5.0:
* the check is missing an `abs()` around values.
* computing the quotient stderr/|value| is problematic when the values are zero. An equivalent check would be stderr < threshold * |value|.
* The real issue is that this check allows for greater errors when the values are greater, making convergence for most values much harder than for those at the extremes. This probably results in lower rank correlations with true values for lowest and highest values.

mdbenito:
### Description

This PR closes #299.

### Changes

- Substitutes a comparison $a<\epsilon$ for the bogus one $a/b < \epsilon$ (absolute threshold)
- Fixes bug in `n_jobs` for `truncatd_montecarlo_shapley`
- Fixes composition of converged values from stopping criteria

### Checklist

- [x] Updated Changelog


Xuzzo (approved):
Looks good to me

PR #305, opened 2023-03-08
Add burn-in param to AbsoluteStandardError

mdbenito:
### Description

This PR fixes a problem with `AbsoluteStandardError` whereby empty results are marked as having converged because an empty result has zero variance.

### Changes

- Add a burn-in parameter to allow for the variances to have been updated to a realistic value. 

### Checklist

- [x] Updated Changelog

### DIFF
--- a/src/pydvl/value/stopping.py
+++ b/src/pydvl/value/stopping.py
@@ -192,17 +192,28 @@ class AbsoluteStandardError(StoppingCriterion):
         the marginal contribution formula).
     :param fraction: The fraction of values that must have converged for the
         criterion to return :attr:`~pydvl.utils.status.Status.Converged`.
+    :param burn_in: The number of iterations to ignore before checking for
+        convergence. This is required because computations typically start with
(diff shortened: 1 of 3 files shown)


mdbenito:
@AnesBenmerzoug Tagging you so that you also see the ugliness here (Fabio already has). I'm not very proud of this tbh.

src/pydvl/value/stopping.py
```diff
@@ -192,17 +192,28 @@ class AbsoluteStandardError(StoppingCriterion):
         the marginal contribution formula).
     :param fraction: The fraction of values that must have converged for the
         criterion to return :attr:`~pydvl.utils.status.Status.Converged`.
+    :param burn_in: The number of iterations to ignore before checking for
```
AnesBenmerzoug, line 195: I think we should emphasize the arbitrary nature of the default value.

AnesBenmerzoug (approved):
Looks good to me! I just left one comment about the documentation.

PR #319, opened 2023-03-12
Semi-values and samplers

### LINKED ISSUE #121: Introduce concept  of `Sampler` to determine order of subsets

### LINKED ISSUE #224: Implement generalization to semivalues
This allows faster testing of new choices and provides a more consistent interface

### LINKED ISSUE #225: Implement Beta Shapley
As introduced in _Kwon, Yongchan, and James Zou. “Beta Shapley: A Unified and Noise-Reduced Data Valuation Framework for Machine Learning.” In Proceedings of The 25th International Conference on Artificial Intelligence and Statistics, 8780–8802. PMLR, 2022. https://proceedings.mlr.press/v151/kwon22a.html_.

It would be best to make this depend on #224

mdbenito:
### Description

This PR closes #121, closes #224 and closes #225. It doesn't remove any previous functionality and thus could make it into https://github.com/appliedAI-Initiative/pyDVL/milestone/14 . 

Things missing (for other PRs)

- #322 
- #323
- #321 
- #325 

### Changes

- Introduces the abc `PowersetSampler` and several specializations for uniform, antithetic, deterministic, or permutation sampling.
- Introduces semi-values as the combination of a sampler and a `Coefficient`, with only one function allowing for all combinations
- Implements Data Banzhaf and Beta Shapley as particular cases of the above
- Fixes a bug with boolean operations for `StoppingCriterion.completion()`
- Fixes some types here and there
- Fixes notation in the documentation for permutation Shapley

### Checklist

- [x] Updated Changelog


AnesBenmerzoug:
According to [semver.org](https://semver.org/), adding new functionality should not go into a patch version. Only backwards compatible bug fixes should be part of it.

This should be split. The bug fixes can be part of 0.5.1 and the rest can be part of 0.6.0

src/pydvl/utils/numeric.py
```diff
@@ -221,13 +241,29 @@ def linear_regression_analytical_derivative_d_x_d_theta(
...
+    ...
+
+
 def running_moments(
-    previous_avg: FloatOrArray,
-    previous_variance: FloatOrArray,
-    count: IntOrArray,
-    new_value: FloatOrArray,
-) -> Tuple:  # [FloatOrArray, FloatOrArray]:
+    previous_avg: Union[float, "NDArray[np.float_]"],
```
AnesBenmerzoug, line 262: Why not keep the typevar?

src/pydvl/value/semivalues.py
```diff
@@ -0,0 +1,236 @@
...
+        j = k + 1
+        w = B(j + beta - 1, n - j + alpha) / const
+        # return math.comb(n - 1, j - 1) * w * n
+        return float(w)
+
+    return cast(SVCoefficient, beta_coefficient_w)
+
+
+class SemiValueMode(str, Enum):
+    Shapley = "shapley"
```
AnesBenmerzoug, line 183: Isn't this redundant here since we already have other shapley methods?

I understand the reasoning behind adding it here and the fact that it doesn't take effort to do it but I think this should be removed here and documented.

src/pydvl/value/sampler.py
```diff
@@ -0,0 +1,290 @@
...
+           \underset{S \sim \mathcal{D}_{- i}}{\mathbb{E}} [\delta_i (S) c (
+           S)]$$
+
+        We add a factor $c(S_j)$ in order to have this expectation coincide with
+        the desired expression.
+        """
+        ...
+
+
+class DeterministicSampler(PowersetSampler[T]):
```
AnesBenmerzoug, line 188: Perhaps we should alias this as `CombinatorialSampler`?

src/pydvl/utils/numeric.py
mdbenito: Because it would only be a shortcut for the overload, not really a type to be used elsewhere and I preferred the explicit version. The TypeVar should also be called `FloatOrArrayOfFloat` to be precise, which is not really that much shorter anyway

src/pydvl/value/semivalues.py
mdbenito: As discussed in person, and for the record, the idea is to leave this duplicate functionality for a while, while we test it, deprecate the old one and merge everything.

src/pydvl/utils/numeric.py
AnesBenmerzoug: I see. The typevar takes less vertical space and is less confusing so I would still prefer it but I'll leave the decision to you.

src/pydvl/value/sampler.py
AnesBenmerzoug: For the sake of completeness, shouldn't we also create a deterministic sampler for permutation sampling, too?

mdbenito: Done in 2685c969c14b9a73847a317754a260910d76d267

src/pydvl/utils/numeric.py
mdbenito: I compromised with `__future__.annotations` and or'ing types

PR #329, opened 2023-03-18
Fix TMCS starts too many processes and dies

### LINKED ISSUE #292: TMCS starts too many processes (and dies)
With code based off v0.5.0, commit 152ec50820609c75600b1b9e13ad4b0d46954877, ray complains with errors like:

```
2023-02-28 15:45:28,812	WARNING worker.py:1851 -- WARNING: 274 PYTHON worker processes have been started on node: 22020832ef6d4fae8e4bb7b15ae173df3c17d639877d04d413e5c506 with address: 192.168.0.52. This could be a result of using a large number of actors, or due to tasks blocked in ray.get() calls (see https://github.com/ray-project/ray/issues/3644 for some discussion of workarounds).
```

The [issue referenced in the error](https://github.com/ray-project/ray/issues/3644) claims this might be because of nested remote calls.

On my machine, the data utility notebook ends up spitting this out:

```
...
2023-02-28 15:58:12,575	WARNING worker.py:1851 -- WARNING: 676 PYTHON worker processes have been started on node: 22020832ef6d4fae8e4bb7b15ae173df3c17d639877d04d413e5c506 with address: 192.168.0.52. This could be a result of using a large number of actors, or due to tasks blocked in ray.get() calls (see https://github.com/ray-project/ray/issues/3644 for some discussion of workarounds).
(raylet) [2023-02-28 15:58:19,519 E 114931 114931] (raylet) node_manager.cc:3097: 2 Workers (tasks / actors) killed due to memory pressure (OOM), 0 Workers crashed due to other reasons at node (ID: 22020832ef6d4fae8e4bb7b15ae173df3c17d639877d04d413e5c506, IP: 192.168.0.52) over the last time period. To see more information about the Workers killed on this node, use `ray logs raylet.out -ip 192.168.0.52`
(raylet) 
(raylet) Refer to the documentation on how to address the out of memory issue: https://docs.ray.io/en/latest/ray-core/scheduling/ray-oom-prevention.html. Consider provisioning more memory on this node or reducing task parallelism by requesting more CPUs per task. To adjust the kill threshold, set the environment variable `RAY_memory_usage_threshold` when starting Ray. To disable worker killing, set the environment variable `RAY_memory_monitor_refresh_ms` to zero.
2023-02-28 15:58:21,823	WARNING worker.py:1851 -- WARNING: 736 PYTHON worker processes have been started on node: 22020832ef6d4fae8e4bb7b15ae173df3c17d639877d04d413e5c506 with address: 192.168.0.52. This could be a result of using a large number of actors, or due to tasks blocked in ray.get() calls (see https://github.com/ray-project/ray/issues/3644 for some discussion of workarounds).
(raylet) [2023-02-28 15:59:19,521 E 114931 114931] (raylet) node_manager.cc:3097: 4 Workers (tasks / actors) killed due to memory pressure (OOM), 0 Workers crashed due to other reasons at node (ID: 22020832ef6d4fae8e4bb7b15ae173df3c17d639877d04d413e5c506, IP: 192.168.0.52) over the last time period. To see more information about the Workers killed on this node, use `ray logs raylet.out -ip 192.168.0.52`
(raylet) 
(raylet) Refer to the documentation on how to address the out of memory issue: https://docs.ray.io/en/latest/ray-core/scheduling/ray-oom-prevention.html. Consider provisioning more memory on this node or reducing task parallelism by requesting more CPUs per task. To adjust the kill threshold, set the environment variable `RAY_memory_usage_threshold` when starting Ray. To disable worker killing, set the environment variable `RAY_memory_monitor_refresh_ms` to zero.
```

And after a few more iterations, it dies with
```
(raylet) [2023-02-28 16:18:26,974 E 114931 114931] (raylet) worker_pool.cc:524: Some workers of the worker process(953902) have not registered within the timeout. The process is still alive, probably it's hanging during start.
(ShapleyWorker pid=900576) E0228 16:18:22.835699587  902741 chttp2_transport.cc:2721]             keepalive_ping_end state error: 0 (expect: 1)
(ShapleyWorker pid=906401) E0228 16:18:26.290218105  909206 chttp2_transport.cc:2721]             keepalive_ping_end state error: 0 (expect: 1)
(ShapleyWorker pid=906390) E0228 16:18:22.540661878  907569 chttp2_transport.cc:2721]             keepalive_ping_end state error: 0 (expect: 1)
(ShapleyWorker pid=912437) E0228 16:18:25.968255772  914199 chttp2_transport.cc:2721]             keepalive_ping_end state error: 0 (expect: 1)
(ShapleyWorker pid=912395) E0228 16:18:30.959742320  913533 chttp2_transport.cc:2721]             keepalive_ping_end state error: 0 (expect: 1)
```

AnesBenmerzoug:
### Description

This PR closes #292

It does so by using an abstraction based on concurrent.futures instead of using actors.

I tried first to use ray queues to avoid passing the coordinator to the workers 
but they also rely on an actor and it kept dying.

### Changes

- Added `RayExecutor` class based concurrent.futures to parallel package.
- Used new concurrent.futures executor abstraction in TMCS.
- Removed abstract and shapley actor modules.
- Moved TMCS tests to separate module.
- Fixed check for number of subsets in Data Utility Learning class.
- Updated Data Utility Learning notebook.
- Updated Shapley Basic Spotify notebook.

**EDIT** More changes:

- Removed `n_concurrent_computations` and used `n_jobs` instead of it to mean the number of tasks to submit before waiting for results.
- Replaced `n_local_workers` in the `ParallelConfig` class with n_workers and use that to set the number of max_workers in the given Executor.
- Added a `__post_init__` method to `ParallelConfig` to make sure that n_workers is None when address is set.
- Added tests specifically for the executor.
- For the 'sequential' parallel backend, use `ThreadPoolExecutor` with `max_workers=1`.

**EDIT 2** More changes:

- Added `n_cpus_per_job` field to `ParallelConfig`.
- Added `cancel_futures_on_exit` boolean parameter to `RayExecutor`.

**EDIT 3** More changes:

- Changed `n_workers `in ParallelConfig to `n_cpus_local` to align more closely with its meaning in ray.
- Remove `n_cpus_per_job` from ParallelConfig and pass it instead as an option to the executor's submit method as part of the kwargs parameter. Otherwise mypy will complain that the method does not have the same signature as the one defined in the base Executor class.
- Used `max_workers` in RayExecutor as the maximum number of submitted jobs. Take its value from `n_jobs` instead of `n_workers` (which was renamed to `n_cpus_local`).
- Added a new variable inside TMCS with a value of  `2 * effective_n_jobs` to represent the total number of submitted jobs, including the jobs that are running.

### Checklist

- [x] Updated Changelog
- [X] If notebooks were added/changed, added boilerplate cells are tagged with `"nbsphinx":"hidden"`


mdbenito (commented):
Besides the cancelling of tasks, this PR has highlighted our inconsistent (and possibly bogus) use of `n_jobs` everywhere. I think we need to fix it

More generally, I think that we are not really using ray as it's supposed to be used. For one, `ParallelConfig.n_local_workers` is used for `num_cpus` in `ray.init()` which does not have the effect we document: instead it's the number of cpus for a "raylet" (which I guess is the number of cpus available to each node), which is fine when starting a local cluster, but probably not when using an existing one.

What do you think about the idea of setting `max_workers` in the parallel config (with it being a noop for ray, or maybe a check against the number of nodes available for a running cluster), and then using `n_jobs` as the number of tasks, and setting `num_cpus` to 1 in the call to `ray.remote()`?

src/pydvl/value/shapley/truncated.py
```diff
@@ -200,51 +225,55 @@ def truncated_montecarlo_shapley(
...
-    # coordinator.add_workers(
-    #     n_workers=n_jobs,
-    #     u=u_id,
-    #     update_period=worker_update_period,
-    #     config=config,
-    #     truncation=truncation,
-    # )
-    #
-    # return coordinator.run(delay=coordinator_update_period)
+    done.modify_result = True
```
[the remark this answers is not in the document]
AnesBenmerzoug: I copied this over from one of the actor classes. I also thought it was weird to do it this way but I had other things to worry about. I will just remove it.

src/pydvl/utils/parallel/futures.py
```diff
@@ -0,0 +1,81 @@
...
+    yield executor
+
+
+class RayExecutor(Executor):
+    def __init__(self, max_workers: int, config: ParallelConfig):
+        if config.backend != "ray":
+            raise ValueError(
+                f"Parallel backend must be set to 'ray' and not {config.backend}"
+            )
+        self.max_workers = max_workers
```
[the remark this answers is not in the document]
AnesBenmerzoug: Nice catch! I will fix this.

src/pydvl/utils/parallel/futures.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    if config.backend == "ray":
+        max_workers = effective_n_jobs(max_workers, config=config)
+        executor = RayExecutor(max_workers, config)
+    else:
+        raise NotImplementedError(f"Unexpected parallel type {config.backend}")
+    yield executor
+
+
+class RayExecutor(Executor):
```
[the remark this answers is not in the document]
AnesBenmerzoug: I will a few things taking into account ideas from the linked PR. In the current implementation we get `map()` for free by inhereting from the base `Executor` class just like it is done for `ThreadPoolExecutor` and `ProcessPoolExecutor`.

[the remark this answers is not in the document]
mdbenito: Note that the number of chunks is a dummy in the base's map. Only the process executor chunkifies.

src/pydvl/value/shapley/truncated.py
```diff
@@ -200,51 +225,55 @@ def truncated_montecarlo_shapley(
...
+        # Initial batch of computations
+        for _ in range(n_concurrent_computations):
+            future = executor.submit(
+                _permutation_montecarlo_one_step,
+                u,
+                truncation,
+                algorithm,
+            )
+            futures.add(future)
+        while futures:
```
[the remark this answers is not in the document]
AnesBenmerzoug: Ray keeps the original object's reference in a custom `object_ref` attribute that it adds to the created future when calling `ref.future()`. Calling `ray.cancel()` on it should work without a problem. We don't need an extra object.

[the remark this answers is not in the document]
mdbenito: Yes, that's what I meant, but wrapping it in a better implementation of Future, for completeness sake. Or just cancelling it directly in this code.

src/pydvl/utils/parallel/futures.py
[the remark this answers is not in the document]
AnesBenmerzoug: Since we rely on ray and don't implement our own workers we cannot control chunking without modifying the function that we're submitting (we could wrap it in another function and submit that instead though but for I think it would add too much complexity).

I added more code to throttle the number of submitted futures.

AnesBenmerzoug:
@mdbenito Here are the latest changes:

- I removed `n_concurrent_computations` and used `n_jobs` instead of it to mean the number of tasks to submit before waiting for results.
- I replaced `n_local_workers` in the `ParallelConfig` class with `n_workers` and use that to set the number of `max_workers` in the given Executor.
- I created a class that runs in a local thread and throttle the number of submitted futures using two queues. This code was heavily inspired by the `ProcessPoorExecutor` and `ThreadPoolExecutor` classes.
- I added tests specifically for the executor.
- For the 'sequential' parallel backend, we use `ThreadPoolExecutor` with max_workers=1.

src/pydvl/utils/parallel/futures/ray.py
```diff
@@ -0,0 +1,324 @@
...
+        config_dict.pop("n_workers")
+        if "address" not in config_dict:
+            config_dict["num_cpus"] = max_workers
+        self.config = config_dict
+        if not ray.is_initialized():
+            ray.init(**self.config)
+
+        self._max_workers = max_workers
+        if self._max_workers is None:
+            self._max_workers = int(ray._private.state.cluster_resources()["CPU"])
```
mdbenito, line 72: If we stick to 1 CPU per job, then:

```suggestion
        self.n_cpus_per_job = 1  # TODO: make configurable
        available_cpus = int(ray._private.state.cluster_resources()["CPU"])
        if max_workers is None:
            max_workers = available_cpus // self.n_cpus_per_job
        elif max_workers*self.n_cpus_per_job > available_cpus:
            logger.warning(f"max_workers*n_cpus_per_job cannot be larger than the available cluster resources ({available_cpus} CPUs)")
            max_workers = available_cpus // self.n_cpus_per_job
        self._max_workers = max_workers
```

And we should probably allow configuration of `n_cpus_per_job` in the config.

src/pydvl/utils/parallel/futures/ray.py
```diff
@@ -0,0 +1,324 @@
...
+
+            logging.debug("Creating future and putting work item in work queue")
+            future: "Future[T]" = Future()
+            w = _WorkItem(future, fn, args, kwargs)
+            self._put_work_item_in_queue(w)
+            # We delay starting the thread until the first call to submit
+            self._start_work_item_manager_thread()
+            return future
+
+    def shutdown(self, wait: bool = True, *, cancel_futures: bool = False) -> None:
```
mdbenito, line 115: Who calls this? How can one set `cancel_futures` to `True`?

src/pydvl/utils/parallel/futures/ray.py
```diff
@@ -0,0 +1,324 @@
...
+        if not ray.is_initialized():
+            ray.init(**self.config)
+
+        self._max_workers = max_workers
+        if self._max_workers is None:
+            self._max_workers = int(ray._private.state.cluster_resources()["CPU"])
+
+        self._broken = False
+        self._shutdown = False
+        self._cancel_pending_futures = False
```
mdbenito, line 76: Shouldn't this be configurable? We need it for TMCS

mdbenito:
> @mdbenito Here are the latest changes:
> 
> * I removed `n_concurrent_computations` and used `n_jobs` instead of it to mean the number of tasks to submit before waiting for results.
> * I replaced `n_local_workers` in the `ParallelConfig` class with `n_workers` and use that to set the number of `max_workers` in the given Executor.

So the idea is to have `n_jobs*n_cpus_per_job <= n_workers <= cluster_size`, where the latter is handled outside the code, correct? Do we really need both `n_workers` and `n_jobs`?

> * I created a class that runs in a local thread and throttle the number of submitted futures using two queues. This code was heavily inspired by the `ProcessPoorExecutor` and `ThreadPoolExecutor` classes.

Silly question: Isn't this duplicating the scheduling functionality in ray? For a given cluster of size N, I would expect ray to be able to receive 3N scheduling requests, queue 2N of them and schedule from the queue as the first N actors complete their work. Or did you add `n_workers` as a way of allowing the user to use just a subset of the cluster, while still asking for `n_jobs=-1` ? Do we need that?

I'm just a bit put off by the large amount of additional code to maintain. After all, with the removal of the communication queues between workers and coordinator, TMCS now basically does a map-reduce operation on an infinite list. Here's some mock/untested code that doesn't require an executor or manager threads and limits computation to at most `n_jobs` concurrent tasks. It is of course much less general than `RayExecutor` and just an idea.

```python
    def map_reduce_infinite(
        self,
        map_fun: MapFunT,
        reduce_fun: ReduceFunT,
        done: InterruptT,
        n_jobs: int | None = None,
        **kwargs,
    ) -> ValuationResult:
        if len(kwargs) == 0:
            raise ValueError("No arguments provided for map_fun")
        for k, v in kwargs.items():
            kwargs[k] = ray.put(v)
        fun = _wrap_func_with_remote_args(map_fun, **kwargs)

        remotes = []
        acc = ValuationResult.empty()
        while not done(acc):
            remotes.extend(
                ray.remote(num_cpus=self.n_cpus_per_job)(fun).remote(**kwargs)
                for _ in range(n_jobs - len(remotes))
            )
            ready_remotes, _ = ray.wait(
                remotes, num_returns=len(remotes), timeout=self.timeout
            )
            if not ready_remotes:
                sleep(1.0)
                continue
            new_results = ray.get(ready_remotes, timeout=0)
            acc = reduce_fun(acc, new_results)
            remotes = [r for r in remotes if r not in ready_remotes]
        for r in remotes:
            ray.cancel(r)
        return acc
```

> * I added tests specifically for the executor.
> * For the 'sequential' parallel backend, we use `ThreadPoolExecutor` with max_workers=1.

PS: I have only given some superficial thought to these things, so I might be overlooking something. Maybe we can have a meeting to discuss without so much typing if you want.

src/pydvl/utils/parallel/futures/ray.py
AnesBenmerzoug: when using a concurrent.futures Executor class as a context manager, `shutdown()` is called automatically with the default arguments when exiting the block.

In order to call `shutdown()` and pass arguments to it we would have to instantiate the Executor normally and call shutdown manually.

AnesBenmerzoug:
@mdbenito let's discuss this during the next meeting.

src/pydvl/utils/parallel/futures/ray.py
mdbenito: Yes, my point was that it shouldn't be internal only, because task cancellation is necessary.

src/pydvl/utils/parallel/futures/ray.py
```diff
@@ -0,0 +1,324 @@
...
+                    # and then cancel their associated futures.
+                    # We empty the pending queue first.
+                    while True:
+                        with self.queue_lock:
+                            try:
+                                work_item = self.pending_queue.get_nowait()
+                            except queue.Empty:
+                                break
+                            if work_item is not None:
+                                work_item.future.cancel()
```
mdbenito, line 311: As mentioned in https://github.com/appliedAI-Initiative/pyDVL/pull/329#discussion_r1141327845, this is probably a noop for ray's returned futures. One must either wrap those to make `cancel()` work, or call `ray.cancel()` on them.

src/pydvl/utils/parallel/futures/ray.py
AnesBenmerzoug: Yes, I understood what you meant. What I said in return was that we don't have to change the method signature, only the way we use the class. i.e. use

```python
executor = RayExecutor()
...
executor.shutdown(cancel_futures=True)
```

instead of

```python
with RayExecutor() as executor:
    ...
```

src/pydvl/utils/parallel/futures/ray.py
AnesBenmerzoug: I changed things a bit in the code. The work items that are still in the queues were not yet submitted to the ray cluster so there is no need for that.

Perhaps I should, like previously, keep the submitted futures in a weakset and call `ray.cancel()` on them.
I will add that.

AnesBenmerzoug: Done.

src/pydvl/utils/parallel/futures/ray.py
mdbenito: 😄 We are going in circles. I also understood you. But it is IMO an anti-pattern. When I say we should make it configurable, I mean:
```python
with RayExecutor(cancel_futures=True) as executor:
    ...
```
And I think that `True` should be the default, and cancel pending and already running futures as soon as the context is exited.

AnesBenmerzoug: I get your point but I would like to stick to the Executor's api. I found a solution that will satisfy us both: I changed the `__exit__()` method to set `cancel_futures=True` when calling the `shutdown()` method.

src/pydvl/utils/parallel/futures/ray.py
```diff
@@ -0,0 +1,324 @@
...
+__all__ = ["RayExecutor"]
+
+T = TypeVar("T")
+
+logger = logging.getLogger(__name__)
+
+
+class RayExecutor(Executor):
+    """Asynchronous executor using Ray that implements the concurrent.futures API.
+
```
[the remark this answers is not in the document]
AnesBenmerzoug: Done.

src/pydvl/utils/parallel/futures/ray.py
AnesBenmerzoug: As discussed during the meeting, I added a `cancel_futures_on_exit` parameter.

src/pydvl/utils/parallel/futures/ray.py
AnesBenmerzoug: I added this now as a field in ParallelConfig.

src/pydvl/value/shapley/truncated.py
```diff
@@ -196,55 +229,57 @@ def truncated_montecarlo_shapley(
...
+                futures, timeout=60, return_when=FIRST_COMPLETED
+            )
+            for future in completed_futures:
+                accumulated_result += future.result()
+                if done(accumulated_result):
+                    break
+            if done(accumulated_result):
+                break
+            # Submit more computations
+            # The goal is to always have `n_concurrent_computations`
```
mdbenito, line 275:
```suggestion
            # The goal is to always have `n_jobs`
```

src/pydvl/value/shapley/truncated.py
```diff
@@ -196,55 +229,57 @@ def truncated_montecarlo_shapley(
...
-    #     truncation=truncation,
-    # )
-    #
-    # return coordinator.run(delay=coordinator_update_period)
+    n_jobs = effective_n_jobs(n_jobs, config)
+
+    algorithm = "truncated_montecarlo_shapley"
+
+    parallel_backend = init_parallel_backend(config)
+    u = parallel_backend.put(u)
```
mdbenito, line 248: Isn't it ugly, revealing that one must "put" some data? Ideally this would be hidden from a user of the parallel backend

tests/utils/conftest.py
```diff
@@ -10,7 +10,7 @@ def parallel_config(request, num_workers):
     if request.param == "sequential":
         yield ParallelConfig(backend=request.param)
     elif request.param == "ray-local":
-        yield ParallelConfig(backend="ray")
+        yield ParallelConfig(backend="ray", n_workers=num_workers)
```
mdbenito, line 13: Again, unsure about the semantics of num_workers being consistent

src/pydvl/utils/parallel/futures/ray.py
```diff
@@ -0,0 +1,343 @@
...
+    def submit(self, fn: Callable[..., T], *args, **kwargs) -> "Future[T]":
+        r"""Submits a callable to be executed with the given arguments.
+
+        Schedules the callable to be executed as fn(\*args, \**kwargs)
+        and returns a Future instance representing the execution of the callable.
+
+        :param fn: Callable.
+        :param args: Positional arguments that will be passed to `fn`.
+        :param kwargs: Keyword arguments that will be passed to `fn`.
```
mdbenito, line 92:
```suggestion
        :param args: Positional arguments that will be passed to ``fn``.
        :param kwargs: Keyword arguments that will be passed to ``fn``.
```

src/pydvl/utils/parallel/backend.py
```diff
@@ -169,9 +171,13 @@ def wrap(self, fun: Callable, **kwargs) -> Callable:
...
-        if len(kwargs) > 1:
-            return ray.remote(**kwargs)(fun).remote  # type: ignore
-        return ray.remote(fun).remote  # type: ignore
+        remote_fn = ray.remote(fun)
+        if len(kwargs) == 0:
+            kwargs["num_cpus"] = self.n_cpus_per_job
+        else:
+            if "num_cpus" not in kwargs:
+                kwargs["num_cpus"] = self.n_cpus_per_job
+        return remote_fn.remote  # type: ignore
```
mdbenito, line 180: This seems to be a mistake? With this change, kwargs are no longer passed to `ray.remote()`
Also, see my comment below about `n_cpus_per_job`, and I don't think it belongs in the instance's state. If it isn't set by the user in the kwargs, that's it.
The previous code takes into account the 2 different ways of calling ray.remote, I think it should be left as it was.
```suggestion
        if len(kwargs) > 1:
            return ray.remote(**kwargs)(fun).remote  # type: ignore
        return ray.remote(fun).remote  # type: ignore
```

mdbenito (commented):
I still think that there are some inconsistencies wrt. `max_workers`. The number of cpus available in the cluster is an external factor over which the code has no effect. So we must ignore that, in particular in `ray.init()`, where `num_cpus` does _not_ refer to the number of cpus used for a local cluster.

`max_workers` could then be used as either of:

1. the maximum number of vCPUs to be used by the executor (num_jobs * num_cpus_per_job)
2. the maximum number of tasks to be run by the executor (so that effective_cpus_used = max_workers * n_cpus_per_job).
We need to fix the names once and for all:

* task = job
* worker = single-core process = CPU

I find the second one horrible, but that seems to be ray's convention, right? We don't have to follow it though: in the `ParallelConfig` and elsewhere we could use `max_cpus` instead of `max_workers`. The question is then what to do when we allow for additional resources like GPUs

src/pydvl/utils/parallel/futures/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+    ...     results = list(executor.map(lambda x: x + 1, range(5)))
+    ...
+    >>> print(results)
+    [1, 2, 3, 4, 5]
+
+    """
+    if config.backend == "ray":
+        max_workers = config.n_workers
+        with RayExecutor(max_workers, config=config) as executor:
+            yield executor
```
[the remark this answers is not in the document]
AnesBenmerzoug: No, it won't. The inner context's, , `RayExecutor`, `__exit__` method will be called first and cancel the futures and then the outer context's, `init_executor`, `__exit__` will be called.

For example, running the following code:

```python
from contextlib import contextmanager

@contextmanager
def func1():
    yield
    print("exit func1")

@contextmanager
def func2():
    with func1():
        yield
    print("exit func2")

with func2():
    pass
```

Will print:

```python
exit func1
exit func2
```

src/pydvl/value/shapley/truncated.py
AnesBenmerzoug: This is the best I could come up with without tying the executor to the specific parallel backend.

src/pydvl/utils/parallel/futures/ray.py
```diff
@@ -0,0 +1,343 @@
...
+            if max_workers <= 0:
+                raise ValueError("max_workers must be greater than 0")
+            max_workers = max_workers
+
+        self.cancel_futures_on_exit = cancel_futures_on_exit
+
+        config_dict = asdict(config)
+        config_dict.pop("backend")
+        config_dict.pop("n_workers")
+        self.n_cpus_per_job = config_dict.pop("n_cpus_per_job")
```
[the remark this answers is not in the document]
AnesBenmerzoug: Sure, I'll remove it but if I remember correctly it was you who suggested adding it to ParallelConfig.

src/pydvl/utils/parallel/futures/ray.py
```diff
@@ -0,0 +1,343 @@
...
+        executor = self.executor_reference()
+        with self.shutdown_lock:
+            logger.debug("work item manager thread acquired shutdown lock")
+            if executor is not None:
+                executor._shutdown = True
+                # Cancel pending work items if requested.
+                if executor._cancel_pending_futures:
+                    logger.debug("forcefully cancelling running futures")
+                    # We cancel the future's object references
+                    # We cannot cancel a running future object.
```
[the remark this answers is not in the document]
AnesBenmerzoug: `ray.cancel(force=True)` cancels the task on the ray cluster but `Future` objects cannot be cancelled if their status is set to `RUNNING`. 

When calling `ray.cancel()` the future will terminate and its result will be an exception.

This is the code of the `cancel()` method:

```python
    def cancel(self):
        """Cancel the future if possible.

        Returns True if the future was cancelled, False otherwise. A future
        cannot be cancelled if it is running or has already completed.
        """
        with self._condition:
            if self._state in [RUNNING, FINISHED]:
                return False

            if self._state in [CANCELLED, CANCELLED_AND_NOTIFIED]:
                return True

            self._state = CANCELLED
            self._condition.notify_all()

        self._invoke_callbacks()
        return True
```

AnesBenmerzoug:
@mdbenito I read the ray documentation and architecture more thoroughly and here's what I found:

- According to [this section](https://docs.ray.io/en/latest/ray-core/scheduling/resources.html#physical-resources-and-logical-resources) of their documentation:

  - Resource requirements of tasks or actors do NOT impose limits on actual physical resource usage. 
  - Ray doesn’t provide CPU isolation for tasks or actors. 

- According to this [other section](https://docs.ray.io/en/latest/ray-core/scheduling/resources.html#specifying-node-resources) of their document:

  - By default, Ray nodes start with pre-defined CPU, GPU, and memory resources. The quantities of these resources on each node are set to the physical quantities auto detected by Ray. By default, logical resources are configured by the following rule:
    - Number of logical CPUs (``num_cpus``): Set to the number of CPUs of the machine/container. 
  - Using ``ray.init()`` to start a single node Ray cluster and setting `num_cpus` will start a Ray node with `num_cpus` logical cpus i.e. `num_cpus` worker processes.

- According to yet [another section](https://docs.ray.io/en/latest/ray-core/scheduling/resources.html#specifying-task-or-actor-resource-requirements) of their documentation:

  - Ray allows specifying a task or actor’s resource requirements (e.g., CPU, GPU, and custom resources). The task or actor will only run on a node if there are enough required resources available to execute the task or actor.

I finally understand this better. Thanks for the link to that document.

Here's what I suggest:

- Change `n_workers` in ParallelConfig to `n_cpus` to align more closely with its meaning in ray.
- Remove `n_cpus_per_job` from ParallelConfig and pass it instead as an option to the executor's constructor. Still I think this needs more thinking.
- Use `max_workers` in RayExecutor as the maximum number of submitted jobs. Take its value from `n_jobs` instead of `n_workers`.
- Add another argument called `batch_size` or something similar to TMCS to represent the number of tasks that will be submitted at each iteration. It can default to `2 * effective_n_jobs`

What do you think?

mdbenito (approved):

PR #338, opened 2023-04-05
Implement class wise shapley

### LINKED ISSUE #259: Implement Class-Wise shapley
As detailed in _Schoch, Stephanie, Haifeng Xu, and Yangfeng Ji. “[CS-Shapley: Class-Wise Shapley Values for Data Valuation in Classification]( https://openreview.net/forum?id=KTOcrOR5mQ9).” 2022_

Also see: https://transferlab.appliedai.de/pills/2022/cs-shapley/

kosmitive:
### Description

This PR closes #259 

### Changes

- Added class wise shapley algorithm
- Added test case for in and out of class accuracy.

### Checklist

- [x] Updated Changelog


src/pydvl/value/shapley/classwise.py
```diff
@@ -0,0 +1,301 @@
+"""
+Refs:
+
+[1] CS-Shapley: Class-wise Shapley Values for Data Valuation in Classification (https://arxiv.org/abs/2211.06800)
```
mdbenito, line 4: We use bibtex for citations. You can see how it's done elsewhere with `:footcite:` . You will need to add the bibliographic data to the bib file in the docs

src/pydvl/value/result.py
```diff
@@ -499,9 +499,9 @@ def __add__(self, other: "ValuationResult") -> "ValuationResult":
 
         """
         # empty results
-        if len(self.values) == 0:
+        if len(self.values) == 0 or np.all(self.counts == 0):
```
mdbenito, line 502: This seems reasonable but is likely to break things. Check the tests for this class and add one for this or the previous behaviour. If it doesn't break anything, then you need to document the behaviour in the docstring.

src/pydvl/value/shapley/classwise.py
```diff
@@ -0,0 +1,301 @@
...
+    "CSScorer",
+]
+
+from sklearn.metrics import accuracy_score
+from tqdm import tqdm
+
+from pydvl.utils.numeric import random_powerset_group_conditional
+from pydvl.value import StoppingCriterion, ValuationResult
+
+IntArray = NDArray[np.int_]
```
mdbenito, line 30: Although convenient, this type definition is not used anywhere else in the code. The inconsistency will make this code harder to read, so I would remove this, or change it consistently everywhere in a separate PR.

src/pydvl/value/shapley/classwise.py
```diff
@@ -0,0 +1,301 @@
...
+    def __init__(self):
+        self.default = 0.0
+        self.range = (0.0, 1.0)
+        self.label: Optional[np.int_] = None
+
+    def __call__(
+        self, model: SupervisedModel, x_test: np.ndarray, y_test: np.ndarray
+    ) -> float:
+        """
+        Estimates the in and out of class accuracies and aggregated them into one float number.
```
mdbenito, line 47:
```suggestion
        Estimates the in- and out-of-class accuracies and aggregates them into one float number.
```

src/pydvl/value/shapley/classwise.py
```diff
@@ -0,0 +1,301 @@
...
+        :param model: A model to be used for predicting the labels.
+        :param x_test: The inputs to be used for measuring the accuracies. Has to match the labels.
+        :param y_test:  The labels ot be used for measuring the accuracies. It is divided further by the passed label.
+        :return: The aggregated number specified by 'in_cls_acc * exp(out_cls_acc)'
```
mdbenito, line 51: Please respect the maximum column width. Sadly, black does not look at comments.

Also, the formulation is a bit unclear. What does "has to match the labels" mean? and what does it mean that y_test is "divided further"? Do you mean a stratified split?

src/pydvl/value/shapley/classwise.py
```diff
@@ -0,0 +1,301 @@
...
+        indices=indices,
+        data_names=u.data.data_names[indices],
+    )
+
+    _, y_train = u.data.get_training_data(indices)
+    unique_labels = np.unique(y_train)
+    pbar = tqdm(disable=not progress, position=0, total=100, unit="%")
+
+    while not done(result):
+
```
mdbenito, line 102: The convention is _not_ to leave blank lines after `while`, `if`, `for`, etc.

src/pydvl/value/shapley/classwise.py
```diff
@@ -0,0 +1,301 @@
...
+                u.model, u.data.x_test, u.data.y_test, label
+            )
+
+            sigma = np.sum(result.values[label_set])
+            result.values[label_set] *= in_cls_acc / sigma
+
+    return result
+
+
+def __class_complement_conditional_sampler(
```
mdbenito, line 175: Double underscores have a special meaning in python. Use just one, or in this case probably none. Also, this function is called `...sampler` but it does some additional computation, so it's a misnomer.

Finally, passing a fixed number of samples is not very robust nor flexible. That's why we have `StoppingCriteria`. You can also use them for an inner loop like this, possibly a separate criterion.

src/pydvl/value/shapley/classwise.py
```diff
@@ -0,0 +1,301 @@
...
+        )
+
+    return result
+
+
+def _estimate_in_out_cls_accuracy(
+    model: SupervisedModel, x: np.ndarray, labels: np.ndarray, label: np.int_
+) -> Tuple[float, float]:
+    """
+    Estimate the in and out of class accuracy as defined in [1], Equation 3.
```
mdbenito, line 220: See my comment above about `:footcite:`

src/pydvl/value/shapley/classwise.py
```diff
@@ -0,0 +1,301 @@
...
+    ):
+        result = __truncated_permutation_mcmc_sampler(
+            u, np.random.permutation(label_set), subset_complement, result, atol=eps
+        )
+
+    return result
+
+
+def _estimate_in_out_cls_accuracy(
+    model: SupervisedModel, x: np.ndarray, labels: np.ndarray, label: np.int_
```
mdbenito, line 217: Use `NDArray[type]` for the types. Also, since you name the data `x`, I think you should name the labels `y`, for consistency

src/pydvl/value/shapley/classwise.py
```diff
@@ -0,0 +1,301 @@
...
+    return label_set, complement_label_set
+
+
+def __truncated_permutation_mcmc_sampler(
+    u: Utility,
+    class_label_set: np.ndarray,
+    complement_set: np.ndarray,
+    result: ValuationResult,
+    atol: float,
+) -> ValuationResult:
```
mdbenito, line 265: As before:
* this is not a "sampler"
* better not to use double underscores
* Use NDArray
* Don't use tolerances, but `StoppingCriterion` or `TruncationPolicy`

src/pydvl/utils/numeric.py
```diff
@@ -113,6 +124,57 @@ def random_powerset(
...
+        raise ValueError("Set and labels have to be of same size.")
+
+    rng = np.random.default_rng()
+    total = 1
+    if n_samples is None:
+        n_samples = np.iinfo(np.int32).max
+
+    unique_labels = np.unique(labels)
+    while total <= n_samples:
+
```
mdbenito, line 161: I think we need to remove `n_samples` from all samplers and have infinite generators. After removing the parameter, documentation, etc.:

```suggestion
    while True:
```
(also, no blank line after `while`)

src/pydvl/utils/utility.py
```diff
@@ -110,7 +110,7 @@ def __init__(
         self,
         model: SupervisedModel,
         data: Dataset,
-        scorer: Optional[Union[str, Scorer]] = None,
+        scorer: Optional[Union[str, ScorerCallable]] = None,
```
mdbenito, line 113: The Scorer implements methods, like `__repr__` that are relied upon in the code, so we can't just change the type. If this is because of your `CSScorer`, you can wrap it in a `Scorer` with the methods provided, without needing to implement those dunders.
```suggestion
        scorer: Optional[Union[str, Scorer]] = None,
```

mdbenito (commented):
I added a few superficial comments but didn't review in depth. Waiting until you are done. Nice progress!

mdbenito (commented):
Added a few more comments, mostly about conventions. Please also respect the column limit in the docstrings (we use black's 88 column default)

src/pydvl/value/shapley/classwise.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+def _normalize_classwise_shapley_values(
+    result: ValuationResult,
+    u: Utility,
+) -> ValuationResult:
+    """
+    Normalize a valuation result by a factor specific to class-wise shapley. Each value
+    corresponds to a class c and gets normalized by multiplying `in-cls-score / sigma`.
+    See also footcite:t:`schoch_csshapley_2022` for more details.
```
mdbenito, line 192: Please explain the terms that you introduce. What's sigma? Only refer to the paper as a last recourse. Users should have all information in the documentation

src/pydvl/value/shapley/classwise.py
```diff
@@ -0,0 +1,216 @@
...
+    if not dim_correct or not is_integral:
+        raise ValueError(
+            "The supplied dataset has to be a 1-dimensional classification dataset."
+        )
+
+    if not isinstance(u.scorer, ClassWiseScorer):
+        raise ValueError(
+            "Please set a subclass of ClassWiseScorer object as scorer object of the"
+            " utility. See"
+            " scoring argument of Utility."
```
mdbenito, line 181: Bogus formatting
```suggestion
            "Please set a subclass of ClassWiseScorer object as scorer object of the"
            " utility. See the scoring argument of Utility."
```

src/pydvl/value/shapley/classwise.py
```diff
@@ -0,0 +1,216 @@
...
+
+    return result
+
+
+def _check_classwise_shapley_utility(u: Utility):
+    """
+    Checks whether the passed utility object is suitable for computing the class-wise
+    shapley values. Therefore, it is checked that the data passed is a pure
+    classification dataset. If the function passes without check it is considered to be
+    run successful.
```
mdbenito, line 157: English fix and simplification:
```suggestion
    Verifies if the provided utility object can calculate class-wise Shapley values.
    This function ensures the input data is strictly for classification purposes. If the
    test is successful, the function runs without issues.
```

The text above is the output of GPT-4 with the prompt:
> Please fix the English usage and simplify as much as possible this text, using language suitable for software documentation

src/pydvl/utils/score.py
```diff
@@ -81,6 +87,153 @@ def __repr__(self):
...
+
+        :param model: Model to be used for determining the score on the validation set.
+        :param x_test: Array containing the features of the associated classification
+            problem.
+        :param y_test: Array containing the labels of the associated classification
+            problem.
+        :param rescale_scores: True if the scores shall be denormalized. This makes
+            sense when for the inner score an estimator of the form 1/N sum_i x_i has
+            been used.
+        :return: Tuple containing the in-cls and out-of-cls score.
```
mdbenito, line 216: Please review this docstring (and all others) for grammar, spelling and other errors. Also, explain all notation properly. Same for docs!

src/pydvl/value/shapley/montecarlo.py
```diff
@@ -152,6 +149,117 @@ def permutation_montecarlo_shapley(
...
+    permutation: NDArray[np.int_],
+    *,
+    truncation: TruncationPolicy,
+    algorithm_name: str,
+    conditional_indices: Optional[NDArray[np.int_]] = None,
+) -> ValuationResult:
+    """
+    A truncated version of a permutation-based MC estimator for class-wise shapley
+    values. It generates a permutation p[i] of the class label indices and iterates over
+    the subsets starting from the set containing only one element up to the full set of
```
mdbenito, line 216: If this statement is true, then the function needs to be changed, as we discussed: the first marginal must be between the empty set and the set containing the point of interest (but yes, this is basically measuring the noise of the estimator with random parameters, because training on one point is meaningless, so it is up to discussion whether we should change this for all methods, see #346 )

mdbenito (commented):
I left more comments. There is so much "peripheral" stuff that I find it hard to focus on the core functionality. Please review everything for spelling, correctness and consistency before requesting another review.

src/pydvl/utils/config.py
```diff
@@ -25,6 +25,7 @@ class ParallelConfig:
     address: Optional[Union[str, Tuple[str, int]]] = None
     n_cpus_local: Optional[int] = None
     logging_level: int = logging.WARNING
+    _temp_dir: Optional[str] = None
```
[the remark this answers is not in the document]
kosmitive: @AnesBenmerzoug It is used to redirect the temp directory to the raid on the DGX.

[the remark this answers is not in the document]
AnesBenmerzoug: Where is it used though? Is it passed through dict unpacking to `ray.init`?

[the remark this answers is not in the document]
kosmitive: Excatly

tests/utils/conftest.py
```diff
@@ -23,3 +27,21 @@ def parallel_config(request, num_workers):
...
+def alternating_sequence_dataset(
+    request,
+) -> Tuple[NDArray[np.float_], NDArray[np.int_], Dict[str, float]]:
+    """
+    The label set is given in the form of 0000011100011111, where the left and right
+    margin are variable. The accuracy can then be determined by a closed-form solution.
+    """
+    n_element = request.param["n_element"]
+    left_margin = request.param["left_margin"]
+    right_margin = request.param["right_margin"]
```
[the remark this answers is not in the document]
kosmitive: Done like in `test_utility`

src/pydvl/value/shapley/montecarlo.py
kosmitive: Done as discussed from first point. Now it matches description.

tests/utils/test_numeric.py
```diff
@@ -138,3 +139,29 @@ def test_running_moments():
...
+
+@pytest.mark.parametrize(
+    "min_elements",
+    [1, 2],
+)
+def test_random_powerset_group_conditional(
+    min_elements: int,
+    elements_per_group: int = 10,
+    num_groups: int = 3,
+    check_num_samples: int = 10,
```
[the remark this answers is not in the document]
kosmitive: Done

tests/misc.py
mdbenito, line 1: These supporting classes typically go into the corresponding `conftest.py`, or even into the file using them if they are local to one test file.

tests/misc.py
```diff
@@ -0,0 +1,36 @@
+import numpy as np
+from numpy._typing import NDArray
```
mdbenito, line 2:
```suggestion
from numpy.typing import NDArray
```

tests/utils/test_numeric.py
```diff
@@ -138,3 +139,29 @@ def test_running_moments():
         true_variances = [np.var(vv) for vv in values]
         assert np.allclose(means, true_means)
         assert np.allclose(variances, true_variances)
+
+
+@pytest.mark.parametrize("min_elements", [1, 2])
+@pytest.mark.parametrize("elements_per_group", [10])
+@pytest.mark.parametrize("num_groups", [3])
+@pytest.mark.parametrize("check_num_samples", [10])
```
mdbenito, line 147: Why three parametrizations instead of a single call? You are forming the cartesian product of singletons...

tests/utils/test_score.py
```diff
@@ -69,3 +81,115 @@ def test_squashed_variance():
     X = np.array([[1, 2], [3, 4]])
     model = FittedLinearModel(coef)
     assert sigmoid(1.0) == squashed_variance(model, X, X @ coef)
+
+
+@pytest.mark.parametrize(
+    "dataset_alt_seq_simple",
+    [((101, 0.3, 0.4))],
+    indirect=True,
```
mdbenito, line 89: This is very opaque. Can't you forward arguments to the fixture like we do elsewhere to expose all the config in the calling test?

src/pydvl/value/result.py
```diff
@@ -234,8 +235,12 @@ def __init__(
 
         self._algorithm = algorithm
         self._status = Status(status)  # Just in case we are given a string
-        self._values = values
-        self._variances = np.zeros_like(values) if variances is None else variances
+        self._values = arr_or_writeable_copy(values)
```
mdbenito, line 238: Where have you encountered a problem with a non-writeable array? Also, the semantics now become a bit obscure. Copy or reuse? I could understand "always copy", or "always reuse", but mixing them can lead to errors...

src/pydvl/value/shapley/classwise.py
```diff
@@ -0,0 +1,251 @@
...
+        function should be of type :class:`~pydvl.utils.score.ClassWiseScorer`.
+    :param done: Function that checks whether the computation needs to stop.
+    :param truncation: Callable function that decides whether to interrupt processing a
+        permutation and set subsequent marginals to zero.
+    :param normalize_values: Indicates whether to normalize the values by the variation
+        in each class times their in-class accuracy.
+    :param n_resample_complement_sets: Number of times to resample the complement set
+        for each permutation.
+    :param use_default_scorer_value: Use default scorer value even if additional_indices
+        is not None.
```
mdbenito, line 63: This needs a lot more explanation 😅

src/pydvl/value/shapley/classwise.py
```diff
@@ -0,0 +1,251 @@
...
+    use_default_scorer_value: bool = True,
+    min_elements_per_label: int = 1,
+    n_jobs: int = 1,
+    config: ParallelConfig = ParallelConfig(),
+    progress: bool = False,
+) -> ValuationResult:
+    """
+    Computes the classwise Shapley value by parallel processing. Independent workers
+    are spawned to process the data in parallel. Once the data is aggregated, the values
+    can be optionally normalized, depending on ``normalize_values``.
```
mdbenito, line 51: The implementation details do not belong in the docsting. Instead we always document what the method computes.

src/pydvl/value/shapley/classwise.py
```diff
@@ -0,0 +1,251 @@
...
+        in each class times their in-class accuracy.
+    :param n_resample_complement_sets: Number of times to resample the complement set
+        for each permutation.
+    :param use_default_scorer_value: Use default scorer value even if additional_indices
+        is not None.
+    :param min_elements_per_label: The minimum number of elements for each opposite
+        label.
+    :param n_jobs: Number of parallel jobs to run.
+    :param config: Parallel configuration.
+    :param progress: Whether to display progress bars for each job.
```
mdbenito, line 68:
```suggestion
    :param progress: Whether to display a progress bar.
```

src/pydvl/value/shapley/montecarlo.py
```diff
@@ -152,6 +148,146 @@ def permutation_montecarlo_shapley(
     return map_reduce_job()
 
 
+def permutation_montecarlo_classwise_shapley(
```
mdbenito, line 151: I would move this function to `classwise.py` and make it private. Why would a user need it? And if it is indeed needed then the name has to change because this is the same as that of functions that compute actual shapley values, whereas this is conditioned on a class. At any rate, this belongs in `classwise.py`

src/pydvl/value/shapley/montecarlo.py
```diff
@@ -152,6 +148,146 @@ def permutation_montecarlo_shapley(
...
+    ):
+        result += _permutation_montecarlo_shapley_rollout(
+            u,
+            indices_permutation,
+            additional_indices=subset_complement,
+            truncation=truncation,
+            algorithm_name=algorithm_name,
+            use_default_scorer_value=use_default_scorer_value,
+        )
+        if done(result):
```
mdbenito, line 208: Is checking here (in the inner, class-conditional loop) the right thing? Values undergo further scaling and weighting after this, correct?

src/pydvl/utils/config.py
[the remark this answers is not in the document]
mdbenito: Unpacking args to ray.init directly is not ideal. As a matter of fact this is no longer how it's done. `RayParallelBackend` extracts the variables it needs explicitly in its constructor, so you would have to change that as well. I would use the opportunity to remove the underscore from this one.

mdbenito (commented):
Thanks for all the work. Now that you seem to have ironed out most bugs, I looked at this again. I left a few general comments, but haven't checked the algorithm itself. Let me know when these are addressed.

src/pydvl/utils/config.py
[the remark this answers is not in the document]
kosmitive: I would actually remove that parameter prior to merging. Reason was to use ray on /raid/. So keep (and rename) or drop?

src/pydvl/utils/dataset.py
```diff
@@ -222,6 +222,10 @@ def indices(self):
         """
         return self._indices
 
+    @indices.setter
```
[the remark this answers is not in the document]
kosmitive: Please review new approach without setter.

src/pydvl/value/shapley/truncated.py
```diff
@@ -111,14 +112,18 @@ class RelativeTruncation(TruncationPolicy):
...
-        self.total_utility = u(u.data.indices)
+        self.total_utility = self.reset(u)
+        self._u = u
 
     def _check(self, idx: int, score: float) -> bool:
         return np.allclose(score, self.total_utility, rtol=self.rtol)
 
-    def reset(self):
-        pass
+    def reset(self, u: Optional[Utility] = None) -> float:
```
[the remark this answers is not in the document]
kosmitive: Some historic artifact. Removed. Confirmed.

src/pydvl/value/shapley/montecarlo.py
kosmitive: It is correct. The done criterion for this specific function is actually not the outer one. It got replaced by `MaxChecks(n_resample_complement_sets)`. To ensure that exactly `n_resample_complement_sets` complement sets are sampled in the function, this call is necessary.

As it wasn't directly clear, do you think it is better to remove the done parameter and replace by the int value directly. And subsequently replace the check with `idx==n_resample_complement_sets`.

src/pydvl/utils/config.py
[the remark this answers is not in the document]
kosmitive: Dropped :)

src/pydvl/value/shapley/classwise.py
```diff
@@ -0,0 +1,575 @@
...
+            done=done_sample_complements,
+            truncation=truncation,
+            use_default_scorer_value=use_default_scorer_value,
+            min_elements_per_label=min_elements_per_label,
+        )
+
+    return result
+
+
+def _check_classwise_shapley_utility(u: Utility):
```
mdbenito, line 195: A function which is used just once is premature generalization. Move into the body of the calling function.

src/pydvl/value/shapley/classwise.py
```diff
@@ -0,0 +1,575 @@
...
+    Verifies if the provided utility object supports classwise Shapley values.
+
+    Args:
+        u: Utility object containing model, data, and scoring function. The scoring
+            function should be of type :class:`~pydvl.utils.score.ClassWiseScorer`.
+
+    Raises:
+        ValueError: If ``u.data`` is not a classification problem.
+        ValueError: If ``u.scorer`` is not an instance of
+            :class:`~pydvl.utils.score.ClassWiseScorer`
```
mdbenito, line 206: Wrong format

src/pydvl/value/shapley/classwise.py
```diff
@@ -0,0 +1,575 @@
...
+def _normalize_classwise_shapley_values(
+    result: ValuationResult,
+    u: Utility,
+) -> ValuationResult:
+    """
+    Normalize a valuation result specific to classwise Shapley.
+
+    Each value corresponds to a class c and gets normalized by multiplying
+    `in-class-score / sigma`. In this context `sigma` is the magnitude of all values
+    belonging to the currently viewed class. See footcite:t:`schoch_csshapley_2022` for
```
mdbenito, line 236: Wrong format, see the example above

src/pydvl/value/shapley/classwise.py
```diff
@@ -0,0 +1,575 @@
...
+    more details.
+
+    Args:
+        result: ValuationResult object to be normalized.
+        u: Utility object containing model, data, and scoring function. The scoring
+            function should be of type :class:`~pydvl.utils.score.ClassWiseScorer`.
```
mdbenito, line 242: Wrong format

src/pydvl/value/shapley/classwise.py
```diff
@@ -0,0 +1,575 @@
...
+    for each element of one class is conditioned on the out-of-cls instances (or a
+    subset of it). The form of the value function can be written as
+
+    .. math::
+        v_{y_i}(D) = f(a_S(D_{y_i}))) * g(a_S(D_{-y_i})))
+
+    where f and g are continuous, monotonic functions and D is the test set.
+
+    in order to produce meaningful results. For further reference see also section four
+    of :footcite:t:`schoch_csshapley_2022`.
```
mdbenito, line 285: Wrong format in 3+ places. Also leftover sentence

src/pydvl/value/shapley/classwise.py
```diff
@@ -0,0 +1,575 @@
...
+            use this to estimate the number of samples required for a certain quality of
+            approximation. If not provided, it can be read from the ``scoring`` object
+            if it provides it, for instance if it was constructed with
+            :func:`~pydvl.utils.types.compose_score`.
+        in_class_discount_fn: Continuous, monotonic increasing function used to
+            discount the in-class score.
+        out_of_class_discount_fn: Continuous, monotonic increasing function used to
+            discount the out-of-class score.
+        initial_label: Set initial label (Doesn't require to set parameter ``label``
+            on ``ClassWiseDiscountedScorer`` in first iteration)
```
mdbenito, line 300: Wrong format

mdbenito (commented):
@kosmitive I fixed the conflicts and undid some unnecessary modifications to code outside `classwise.py`. Still I cannot merge this as it is

* You use what should be precise terms loosely, costring great effort to the reader.
* The tests are almost unintelligible. For example, what is `test_cs_scorer_on_alt_seq_cf_linear_classifier_cs_score`? What are all the others?
* Each test does too many things.
* The huge docstring in `linear_classifier_cs_scorer_args_exact_solution_use_default_score()` belongs somewhere else. As it is, it is almost impossible to read (and no, the solution is not to change the whole documentation system, although we might consider moving all of CONTIBUTING.md to the docs and then adding something about the tests)
* classwise.py does not use the seeding mechanism
* it also doesn't have a proper description of the algorithm itself
* it looks like there at least one level of indirection too many. Remove unnecessary nested functions 
* You use the wrong docstring format in many, many, many places.

I can't keep delaying the release because of this and it's Saturday evening. This will have to wait until 0.7.1

src/pydvl/utils/numeric.py
```diff
@@ -134,6 +156,71 @@ def random_powerset(
...
+        raise ValueError("Set and labels have to be of same size.")
+
+    if min_elements_per_group < 0:
+        raise ValueError(
+            f"Parameter min_elements={min_elements_per_group} needs to be bigger or equal to 0."
+        )
+
+    if min_elements_per_group == 0:
+        logger.warning(
+            "It is recommended to ensure at least one element of each group is"
```
mdbenito, line 196: Recommended for what? This comment is coupled to cs Shapley, but the function available in a common module

src/pydvl/utils/numeric.py
```diff
@@ -134,6 +156,71 @@ def random_powerset(
...
+            )
+            if subset_length > 0:
+                subsets.append(random_subset_of_size(s[label_indices], subset_length))
+
+        if len(subsets) > 0:
+            subset = np.concatenate(tuple(subsets))
+            rng.shuffle(subset)
+            yield subset
+        else:
+            yield np.array([], dtype=int)
```
mdbenito, line 221: Wrong type. Function is generic

src/pydvl/value/result.py
```diff
@@ -541,10 +541,14 @@ def __add__(self, other: "ValuationResult") -> "ValuationResult":
         xm[other_pos] = other._values
         vm[other_pos] = other._variances
 
+        # np.maximum(1, n + m) covers case n = m = 0 with
```
mdbenito, line 544: Clipped sentence

src/pydvl/value/result.py
```diff
@@ -629,6 +633,17 @@ def update(self, idx: int, new_value: float) -> "ValuationResult":
         )
         return self
 
+    def scale(self, coefficient: float, indices: Optional[NDArray[IndexT]] = None):
```
mdbenito, line 636: It's a factor, not a coefficient.

src/pydvl/value/shapley/truncated.py
```diff
@@ -132,15 +134,18 @@ class RelativeTruncation(TruncationPolicy):
     def __init__(self, u: Utility, rtol: float):
         super().__init__()
         self.rtol = rtol
-        logger.info("Computing total utility for permutation truncation.")
```
mdbenito, line 135: Why did you delete the log info?

src/pydvl/value/shapley/truncated.py
```diff
@@ -178,15 +183,15 @@ def _check(self, idx: int, score: float) -> bool:
...
         self.count = 0
         self.variance = self.mean = 0
 
 
 @deprecated(
     target=True,
-    deprecated_in="0.7.0",
-    remove_in="0.8.0",
+    deprecated_in="0.6.1",
+    remove_in="0.7.0",
```
mdbenito, line 194: Revert

src/pydvl/value/stopping.py
```diff
@@ -520,7 +520,7 @@ def _check(self, r: ValuationResult) -> Status:
             quots = np.divide(diffs, curr[ii], out=diffs, where=curr[ii] != 0)
             # quots holds the quotients when the denominator is non-zero, and
             # the absolute difference, which is just the memory, otherwise.
-            if np.mean(quots) < self.rtol:
+            if len(quots) > 0 and np.mean(quots) < self.rtol:
```
mdbenito, line 523: Good catch

tests/utils/test_numeric.py
```diff
@@ -249,3 +250,29 @@ def test_running_moments():
...
+    check_num_samples: int,
+):
+    s = np.arange(num_groups * elements_per_group)
+    groups = np.arange(num_groups).repeat(elements_per_group)
+
+    for idx, subset in enumerate(
+        random_powerset_group_conditional(s, groups, min_elements)
+    ):
+        assert np.all(np.isin(subset, s))
+        assert np.all(np.unique(groups[subset]) == np.unique(groups))
```
mdbenito, line 272: What if min elements is 0?

tests/utils/test_numeric.py
```diff
@@ -249,3 +250,29 @@ def test_running_moments():
...
+
+@pytest.mark.parametrize(
+    "min_elements,elements_per_group,num_groups,check_num_samples",
+    [(1, 10, 3, 3), (2, 10, 3, 3)],
+)
+def test_random_powerset_group_conditional(
+    min_elements: int,
+    elements_per_group: int,
+    num_groups: int,
+    check_num_samples: int,
```
mdbenito, line 263: Maybe increase this param a bit. It's very cheap

mdbenito (commented):
Yet more comments. Please take your time before asking for another review. This is very time consuming. And I'm not even close to finished

src/pydvl/value/shapley/classwise.py
kosmitive: I wouldn't call it premature generalization but more a helper method (although the place might not be the best). But for consistency I removed the function and placed the checking code directly in the function, because the complete (nested) function definition with a docstring makes the method very big.

mdbenito:
@kosmitive Please do not close a PR with hundreds of comments. And please stop force pushing to the repo. We have a clear policy of NO force pushes, ever. Please close the new PR and keep using this one to track progress.

docs/value/classwise-shapley.md
```diff
@@ -0,0 +1,145 @@
...
+    | CPU            | Tabular   | 2       | 21         | 197       |
+    | Covertype      | Tabular   | 7       | 54         | 1596      |
+    | Phoneme        | Tabular   | 2       | 5          | 1489      |
+    | FMNIST         | Image     | 2       | 32         | 40996     |
+    | CIFAR10        | Image     | 2       | 32         | 40927     |
+    | MNIST (binary) | Image     | 2       | 32         | 554       |
+    | MNIST (multi)  | Image     | 10      | 32         | 554       |
+
+
+### Value transfer
```
mdbenito, line 117: Value transfer is more interesting than simple sample removal for the same model, but since you also have the plots for the same model, I think we should add them here.

docs/value/classwise-shapley.md
```diff
@@ -0,0 +1,145 @@
...
+
+We compare CWS to TMCS, Beta Shapley and LOO on a sample removal task with value
+transfer to another model. Values are computed using a logistic regression
+model, and used to prune the training set for a neural network. Random values
+serve as a baseline. The following plot shows valuation-set accuracy of the
+network on the y-axis, and the number of samples removed on the x-axis.
+
+![Accuracy after sample removal using values transferred from logistic regression
+to an MLP](img/classwise-shapley-weighted-accuracy-drop-logistic-regression-to-mlp.svg)
```
mdbenito, line 126: Instead of redoing the plots to have white ticks, which then don't show up in the enlarged image when clicking on it, we have a CSS class "invertible". Just produce an SVG with transparent background and normal colors (i.e. black axes and so on), or remove the background with inkscape, and add the following:
```suggestion
to an MLP](img/classwise-shapley-weighted-accuracy-drop-logistic-regression-to-mlp.svg){ class=invertible }
```
You can see an example in `the-core.md`

docs/value/classwise-shapley.md
```diff
@@ -0,0 +1,145 @@
...
+
+Samples are removed from high to low valuation order and hence we expect a steep
+decrease in the curve. CWS is competitive with the compared methods. Especially
+in very unbalanced datasets, like `Click`, the performance of CWS seems
+superior. In other datasets, like `Covertype` and `Diabetes` and `MNIST (multi)`
+the performance is on par with TMC. For `MNIST (binary)` and `Phoneme` the
+performance is competitive. We remark that for all valuation methods the
+same number of _evaluations of the marginal utility_ was used.
+
+### Value distribution
```
mdbenito, line 136: If the histograms are to be informative, it can only be when comparing them to the histograms of, say, TMCS or any method. Try adding another method to the plots and discussing the differences.

docs/value/img/classwise-shapley-discounted-utility-function.svg
mdbenito, line 1: Please compress the SVGs using [svgo](https://github.com/svg/svgo). Here's a [web app](https://jakearchibald.github.io/svgomg/) if you don't want to use node

mdbenito (changes requested):
I left a couple of comments but this is almost ready. I will wait until you make the changes then do a final review. Please check that my change to the ClasswiseScorer is correct (how did you test with f1 if the class didn't allow changing the metric?)

docs/value/classwise-shapley.md
kosmitive: Added comparison to TMCS, as both are competitive.

mdbenito:
Merging despite many issues that we must leave for a later release. See DM

PR #339, opened 2023-04-05
Draft: Make influence functions great

### LINKED ISSUE #311: Make IF great: Improve interface and include iterative method
There are several changes that are necessary for IF to be great. In order, the following changes must be implemented:
- [x] Conjugate gradient must be implemented within the framework. This is necessary for managing device computation. In the current setup, torch handles all the differentiation, which typically happens in a gpu, but then for hessian inversion data is offloaded to the cpu, causing massive overhead. 
- [x] Influence factors computation should be done in a separate method, so that the result can be stored and re-used. At the moment, influence factors are calculated internally in compute_influences and are not cached.
- [x] The stochastic Hessian inversion (also known as LISSA, Linear time Stochastic Second-Order Algorithm) also needs to be implemented, both for paper reproductions and because it is very useful for large models. 
- [x] Currently code expects all training and test data to be fully loaded into RAM. Include options for pytorch DataLoaders
- [x] With DataLoaders, it will be important to handle the device properly. Check that everything continues to work on a gpu

For now, all these things should be done in Pytorch, keeping in mind that the library aims to be eventually framework independent.

Xuzzo:
### Description

This PR closes #311 

### Changes

- 

### Checklist

- [ ] Updated Changelog


src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -26,9 +27,164 @@ def flatten_gradient(grad):
...
+    :param maxiter: maximum number of iterations,
+    :param damp: damping factor, defaults to 0 for no damping
+    :param scale: scaling factor, defaults to 10
+    :param progress: If True, display progress bars.
+
+    :return: A matrix of shape [NxP] with each line being a solution of $Ax=b$.
+    """
+    h_estimate = torch.clone(b)
+    for _ in maybe_progress(range(maxiter), progress, desc="Lissa"):
+        x, y = next(iter(training_data))
```
AnesBenmerzoug, line 183: Calling `iter()` on training_data on each iteration recreates the iterator each time.
I think creating it outside the loop would be better `iterator = iter(training_data)`

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -47,6 +203,11 @@ def stack(a: Sequence[torch.Tensor], **kwargs):
     return torch.stack(a, **kwargs)
 
 
+def cat(a: Sequence[torch.Tensor], **kwargs):
```
AnesBenmerzoug, line 206: Is this really needed?

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -26,9 +27,164 @@ def flatten_gradient(grad):
...
+    :param lam: regularization of the hessian
+    :param maxiter: maximum number of iterations,
+    :param damp: damping factor, defaults to 0 for no damping
+    :param scale: scaling factor, defaults to 10
+    :param progress: If True, display progress bars.
+
+    :return: A matrix of shape [NxP] with each line being a solution of $Ax=b$.
+    """
+    h_estimate = torch.clone(b)
+    for _ in maybe_progress(range(maxiter), progress, desc="Lissa"):
```
AnesBenmerzoug, line 182: Perhaps doing this with a check inside the loop would be better for readability?

```python
for i, (x, y) in enumerate(maybe_progress(training_data, progress, desc="Lissa")):
```

```python
    if i >= max_iters:
        break
```

src/pydvl/influence/frameworks/torch_differentiable.py
Xuzzo: it is created (like many of the methods here e.g. stack) to keep general.py framework independent...

AnesBenmerzoug: I see.

How about using [singledispatch](https://docs.python.org/3/library/functools.html#functools.singledispatch) from functools instead?

You could have a generic implementation that does nothing and then specialized implementations for each framework.

For example, in separate module (maybe named `common`) inside the frameworks package we could define:

```python
from functools import singledispatch
from typing import Any

@singledispatch
def cat(a: Any) -> Any:
    """Concatenates a sequence of arrays into array"""
    raise RunetimeError(f"Array of type {type(a)} is not supported")
```

And then inside the torch_differentiable module we would do:

```python
import torch

from .common import cat

@cat.register
def cat(a: torch.Tensor) -> torch.Tensor:
    """Concatenates a sequence of tensors into a single torch tensor"""
    return torch.cat(a, **kwargs)
```

This would simplify a bit the `__init__.py` module of the frameworks package.
You would only need to expose the functions from the common module.

You would still need the import guard from the different framework modules though. So it's not a perfect solution, just a small improvement.

P.S. The code is untested (it should work tough) and is just a suggestion.

Xuzzo: Yeah the __init__ is a mess at the moment, but I am leaving a better handling of it for another time, or I will go crazy. Will open an issue with this suggestion =)

src/pydvl/influence/frameworks/torch_differentiable.py
Xuzzo: What we want to do here is to sample a batch with different random points every time. I searched around and there seems to be no clean way to do it apart for re-creating the iterator every time.

src/pydvl/influence/frameworks/torch_differentiable.py
Xuzzo: Similarly to [here](https://github.com/appliedAI-Initiative/pyDVL/pull/339#discussion_r1159702128): we dont want to sample all the points, but a random batch at every loop

src/pydvl/influence/frameworks/torch_differentiable.py
AnesBenmerzoug: I see. Makes sense. Thanks for the explanation.

src/pydvl/influence/inversion.py
```diff
@@ -31,15 +27,16 @@ class InversionMethod(str, Enum):
...
 
 def solve_hvp(
     inversion_method: InversionMethod,
     model: TwiceDifferentiable[TensorType, ModelType],
-    x: TensorType,
-    y: TensorType,
+    training_data: DataLoaderType,
     b: TensorType,
     lam: float = 0,
+    inversion_method_kwargs: Dict[str, Any] = {},
```
AnesBenmerzoug, line 39: wouldn't using `**kwargs` work as just well and be a bit more readable?

```suggestion
    **kwargs: Any,
```

src/pydvl/influence/inversion.py
```diff
@@ -31,15 +27,16 @@ class InversionMethod(str, Enum):
...
 
 
 def solve_hvp(
     inversion_method: InversionMethod,
     model: TwiceDifferentiable[TensorType, ModelType],
-    x: TensorType,
-    y: TensorType,
+    training_data: DataLoaderType,
     b: TensorType,
     lam: float = 0,
```
AnesBenmerzoug, line 38:
```suggestion
    lam: float = 0.0,
```

AnesBenmerzoug: Just for the sake of readability.

src/pydvl/influence/inversion.py
```diff
@@ -31,15 +27,16 @@ class InversionMethod(str, Enum):
...
+    inversion_method_kwargs: Dict[str, Any] = {},
     progress: bool = False,
```
AnesBenmerzoug, line 40: I think making `progress` a keyword only argument makes sense. It is completely optional and users would have to explicitly set it.

```suggestion
    *,
    progress: bool = False,
```

src/pydvl/influence/inversion.py
```diff
@@ -53,52 +50,41 @@ def solve_hvp(
     :param inversion_method:
     :param model: A model wrapped in the TwiceDifferentiable interface.
     :param x: An array containing the features of the input data points.
+    :param y: labels for x
     :param b:
```
AnesBenmerzoug, line 54: What is `b`?

src/pydvl/influence/general.py
```diff
@@ -125,61 +128,63 @@ def _compute_influences_pert(
...
-    x: TensorType,
-    y: TensorType,
-    x_test: TensorType,
-    y_test: TensorType,
+    training_data: DataLoaderType,
+    test_data: Optional[DataLoaderType] = None,
+    input_data: Optional[DataLoaderType] = None,
     progress: bool = False,
     inversion_method: InversionMethod = InversionMethod.Direct,
+    inversion_method_kwargs: Dict[str, Any] = {},
```
AnesBenmerzoug, line 171: Same comment as below:

```suggestion
    **kwargs: Any,
```

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -19,19 +20,201 @@
...
+def solve_batch_cg(
+    model: TwiceDifferentiable,
+    training_data: DataLoader,
+    b: torch.Tensor,
+    lam: float = 0,
+    x0: Optional[torch.Tensor] = None,
+    rtol: float = 1e-7,
+    atol: float = 1e-7,
+    maxiter: Optional[int] = None,
+    progress: bool = True,
```
AnesBenmerzoug, line 70:
```suggestion
    progress: bool = False,
```

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -19,19 +20,201 @@
...
 
-def solve_linear(matrix: torch.Tensor, b: torch.Tensor):
-    """Computes the solution of a square system of linear equations"""
-    return torch.linalg.solve(matrix, b)
+def solve_linear(
+    model: TwiceDifferentiable,
+    training_data: DataLoader,
+    b: torch.Tensor,
+    lam: float = 0,
+    progress: bool = True,
```
AnesBenmerzoug, line 35:
```suggestion
    progress: bool = False,
```

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -19,19 +20,201 @@
...
+    :param lam: regularization of the hessian
+    :param progress: If True, display progress bars.
 
-def as_tensor(a: Any, warn=True, **kwargs):
+    :return: An array that solves the inverse problem,
+        i.e. it returns $x$ such that $Hx = b$
+    """
+    all_x, all_y = [], []
+    for x, y in training_data:
+        all_x.append(x)
```
AnesBenmerzoug, line 51: Isn't there a risk to run out of memory doing it this way?

Xuzzo: yes. That is why the direct method can be used only with tiny models and training sets and typically approximate methods are used

Xuzzo: There is one alternative, which is to keep one grad per batch, like it is done for cg. Maybe we can discuss what makes more sense.

tests/influence/test_influences.py
```diff
@@ -76,52 +77,24 @@ def analytical_linear_influences(
...
+    "influence_type",
+    InfluenceType,
+    ids=[ifl.value for ifl in InfluenceType],
+)
+@pytest.mark.parametrize(
+    "train_set_size",
+    [200],
+    ids=["train_set_size_200"],
 )
 def test_influence_linear_model(
```
kosmitive, line 91: Probably better to split up in different test cases or using parameterization.

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -2,36 +2,231 @@
...
+    :return: A matrix of shape [NxP] with each line being a solution of $Ax=b$.
+    """
+    if h0 is None:
+        h_estimate = torch.clone(b)
+    else:
+        h_estimate = h0
+    shuffled_training_data = DataLoader(
+        training_data.dataset, training_data.batch_size, shuffle=True
+    )
+    lissa_step = lambda h: b + (1 - damp) * h - reg_hvp(h) / scale  # type: ignore
```
AnesBenmerzoug, line 210: Could you please make this a proper function with a docstring? It would make the code more readable

AnesBenmerzoug (approved):
It looks good to me! I just left one comment about making the `lissa_step` lambda into a proper function, but other than that it should be ready to be merged.

src/pydvl/influence/frameworks/__init__.py
```diff
@@ -15,14 +15,17 @@
...
         solve_linear,
+        solve_lissa,
         stack,
     )
 
     TensorType = torch.Tensor
+    DataLoaderType = torch.utils.data.DataLoader
     ModelType = torch.nn.Module
 
 except ImportError:
```
mdbenito, line 31: Won't the extension of `__all__` be bogus if the import fails? We are declaring the module to export symbols that are not there. Depending on what other parts of the influence module expects, one should probably fail here with a warning (until there are other frameworks). Or does this break things elsewhere?

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -2,36 +2,231 @@
...
 
-def solve_linear(matrix: torch.Tensor, b: torch.Tensor):
-    """Computes the solution of a square system of linear equations"""
-    return torch.linalg.solve(matrix, b)
+def solve_linear(
+    model: TwiceDifferentiable,
+    training_data: DataLoader,
+    b: torch.Tensor,
+    *,
+    lam: float = 0.0,
```
mdbenito, line 43: `lam` seems a bit unintuitive for a parameter name. Why not `hessian_perturbation` or `regularizer`. I'd prefer the former because it's explicit about the type of regularization. Note that this should be added to the documentation, since it's $(H + \lambda I) x = b$ that is solved.

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -2,36 +2,231 @@
...
+    )
+    lissa_step = lambda h: b + (1 - damp) * h - reg_hvp(h) / scale  # type: ignore
+    for _ in maybe_progress(range(maxiter), progress, desc="Lissa"):
+        x, y = next(iter(shuffled_training_data))
+        grad_xy, _ = model.grad(x, y)
+        reg_hvp = lambda v: mvp(grad_xy, v, model.parameters) + lam * v
+        residual = lissa_step(h_estimate) - h_estimate
+        h_estimate += residual
+        if torch.isnan(h_estimate).any():
+            raise RuntimeError("NaNs in h_estimate. Increase scale or damp.")
```
mdbenito, line 218:
```suggestion
            raise RuntimeError("NaNs in h_estimate. Increase scale or dampening.")
```

mdbenito (commented):
This looks much better with the dataloaders! I've looked at half of the files and left some comments, mostly in a puny attempt to create the illusion of adding value 😄.

tests/influence/test_influences.py
Xuzzo: I have refactored it a bit. Hopefully it is better.

src/pydvl/influence/frameworks/torch_differentiable.py
Xuzzo: Should be done. Let me know if you feel it needs more docs

src/pydvl/influence/frameworks/__init__.py
Xuzzo: Should be done

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -2,36 +2,231 @@
 Contains all parts of pyTorch based machine learning model.
```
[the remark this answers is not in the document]
Xuzzo: Should be better now.

notebooks/support/torch.py
```diff
@@ -90,43 +90,33 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
-    x_train: torch.Tensor,
-    y_train: torch.Tensor,
-    x_val: torch.Tensor,
-    y_val: torch.Tensor,
+    training_data: DataLoader,
+    val_data: DataLoader,
     loss: Callable[[torch.Tensor, torch.Tensor], torch.Tensor],
     optimizer: Optimizer,
     scheduler: Optional[_LRScheduler] = None,
     num_epochs: int = 1,
```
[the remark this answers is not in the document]
Xuzzo: New issue: #351

tests/influence/test_influences.py
```diff
@@ -142,29 +128,33 @@ def test_influence_linear_model(
...
-        *test_data,
-        progress=True,
-        influence_type=influence_type,
-        inversion_method="direct",
-        hessian_regularization=hessian_reg,
-    ).numpy()
+    train_data_loader = DataLoader(list(zip(*train_data)), batch_size=40, shuffle=True)
+    input_data = DataLoader(list(zip(*train_data)), batch_size=40)
+    test_data_loader = DataLoader(
+        list(zip(*test_data)),
```
mdbenito, line 134: Doesn't building a list defeat the purpose of using generators?

tests/influence/test_influences.py
```diff
@@ -187,94 +177,107 @@ def test_influence_linear_model(
...
+        "direct": {},
+        "cg": {},
+        "lissa": {
+            "maxiter": 100,
+            "scale": 10000,
+        },
+    }
+    train_data_loader = DataLoader(list(zip(x_train, y_train)), batch_size=batch_size)
+    test_data_loader = DataLoader(
+        list(zip(x_test, y_test)),
```
mdbenito, line 252: Same comment: I don't think one should explicitly instantiate a list

mdbenito (commented):
I left another couple of comments, on `list` and `DataLoader`

mdbenito (commented):
I left another couple of comments, on `list` and `DataLoader`

tests/influence/test_influences.py
Xuzzo: I have replaced this ugly list(zip(...)) thingy with TensorDataset when the data is already type Tensor. In the cases (like in this test) where the input is a np.ndarray I have left the syntax since the conversion (copying) needs to be done anyway.

mdbenito: If you create a list from an ndarray, and pass it to the constructor of the data loader, you are potentially copying data twice, depending on how DataLoader is implemented. If it just stores a reference to the argument, you are ok, but if it copies the data into some internal structure, then... double copy.

Also, DataLoader supports iterable datasets for streaming data, so it should be possible to avoid the extra list construction.

Of course none of this matters in a test with a few dozen samples.

PR #342, opened 2023-04-16
Fix publish CI pipeline

AnesBenmerzoug:
### Description

This PR is the second attempt at fixing the publish CI pipeline. 

It wasn't triggering automatically so after doing some research I added a branches filter to the trigger.

I also fixed the check for when we're using [act](https://github.com/nektos/act) to run the pipeline locally.

### Changes

- Added branches filter to publish CI pipeline.
- Used `env.ACT` instead of `github.event.act` as variable to fail in case we're running the CI pipeline locally.

### Checklist

- [ ] Updated Changelog

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -2,6 +2,8 @@ name: Publish Python Package to PyPI
 
 on:
   push:
+    branches:
+    - "**"
     tags:
     - "v*"
   workflow_dispatch:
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -397,8 +397,11 @@ If running in interactive mode (without `-y|--yes`), the script will output a
 summary of pending changes and ask for confirmation before executing the
 actions.
 
-Once this is done, a package will be automatically created and published from CI
-to PyPI.
+Once this is done, a tag will be created on the repository. 
+You should then create a Github
```
mdbenito, line 401: What about [creating the release](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release) as a draft, so that we need only edit it a bit? It's a nice touch with one curl invocation :)

mdbenito (approved):

PR #355, opened 2023-05-29
Add joblib as backend.

### LINKED ISSUE #353: Write joblib backend
As the ray backend can be difficult to configure, we should implement a joblib one for single-machine use. This would just be a subclass of `ParallelBackend`

kosmitive:
### Description

This PR closes #353 

### Changes

- Add `joblib` as a parallel backend.
- Use loky executor for `backend=="joblib"` for futures interface.
- Use joblib directly inside `MapReduceJob` instead of the different parallel backend classes.
- Remove dummy 'sequential' parallel backend.

### Checklist

- [X] Updated Changelog


Xuzzo (approved):
Looks good to me.

kosmitive:
Sequential backend not needed for debugging?

AnesBenmerzoug:
@kosmitive Setting `n_jobs=1` when using joblib's `Parallel` interface disables parallelism and runs everything in the current thread so there was no need for a separate `sequential` backend.

PR #372, opened 2023-07-02
Fix parallelization of permutation sampling

### LINKED ISSUE #364: Bogus handling of termination in PMCS
We cannot use `MapReduce` with permutation sampling because the stopping criterion is not applied to the aggregated result.
- [x] Implement permutation montecarlo using the same method as truncated (merging both)
- [x] Same problem happens with `semivalues`

mdbenito:
### Description

This PR closes #364 and does some cleanup in the semivalues module by moving the sampling to the main process and dispatching single samples to the worker pool. This solves the problem with the stopping criteria. However, the general approach with semi-values is somewhat rigid and TruncationPolicies remain unimplemented.

### Changes

- Adds `Executor` factories to the parallel backends
- Moves the backend registry to `BaseParallelBackend`
- Moves `NoPublicConstructor` to `types.py` where it belongs
- Deprecates the interface `truncated_montecarlo_shapley` in favour of a single `permutation_montecarlo_shapley`
- Changes `permutation_montecarlo_shapley` not to use MapReduce but the futures mechanism
- Fixes generic semi-value computation, changing the way tasks are submitted.
- Aliases all methods to point to the same implementation
- Deprecates the common interface `compute_semivalues` in favour of more explicit wrappers around the common methods
- Fixes a bug in sampler slicing, whereby the wrong index set was being used. Note: with the current strategy, slicing is no longer necessary and we could remove this functionality altogether.
- Changes the tests for valuation methods to only use the joblib backend to save time
- Tweaks a couple of test parameters to make them faster and less finecky.
- Updates and slightly improves some docs

### Checklist

- [x] Updated Changelog


src/pydvl/utils/parallel/futures/ray.py
```diff
@@ -120,12 +132,29 @@ def submit(self, fn: Callable[..., T], *args, **kwargs) -> "Future[T]":
             self._start_work_item_manager_thread()
             return future
 
-    def shutdown(self, wait: bool = True, *, cancel_futures: bool = False) -> None:
+    def shutdown(self, wait: bool = True, *, cancel_futures: bool = None) -> None:
```
AnesBenmerzoug, line 135:
```suggestion
    def shutdown(self, wait: bool = True, *, cancel_futures: Optional[bool] = None) -> None:
```

src/pydvl/utils/parallel/futures/ray.py
```diff
@@ -323,13 +347,18 @@ def flag_executor_shutting_down(self):
...
                 executor._shutdown = True
-                # Cancel pending work items if requested.
-                if executor._cancel_pending_futures:
+                if executor._cancel_futures & CancellationPolicy.RUNNING:
                     logger.debug("forcefully cancelling running futures")
                     # We cancel the future's object references
                     # We cannot cancel a running future object.
                     for future in self.submitted_futures:
                         ray.cancel(future.object_ref)
+                    # FIXME: huh? why is this here?
```
AnesBenmerzoug, line 356: I just copied it from the [original code](https://github.com/python/cpython/blob/3.11/Lib/concurrent/futures/process.py#L529). It probably prevents rerunning this loop in some edge cases.

src/pydvl/utils/parallel/futures/ray.py
```diff
@@ -323,13 +347,18 @@ def flag_executor_shutting_down(self):
...
+                    # Make sure we do this only once to not waste time looping
+                    # on running processes over and over.
+                    executor._cancel_futures = CancellationPolicy.NONE
+
+                if executor._cancel_futures & CancellationPolicy.PENDING:
```
AnesBenmerzoug, line 361: I think this block should run first to drain the pending and work queues to prevent submitting any further jobs and then we can cancel the running.

src/pydvl/value/sampler.py
```diff
@@ -51,22 +55,35 @@
 ]
 
 T = TypeVar("T", bound=np.generic)
-SampleType = Tuple[T, NDArray[T]]
+SampleT = Tuple[T, NDArray[T]]
```
AnesBenmerzoug, line 58: I think `SampleType` was easier to understand.

tests/value/conftest.py
```diff
@@ -123,16 +123,6 @@ def linear_shapley(linear_dataset, scorer, n_jobs):
...
-        ray.shutdown()
-    elif request.param == "ray-external":
-        # Starts a head-node for the cluster.
-        cluster = Cluster(initialize_head=True, head_node_args={"num_cpus": 4})
-        yield ParallelConfig(backend="ray", address=cluster.address)
-        ray.shutdown()
-        cluster.shutdown()
+@pytest.fixture(scope="module")
+def parallel_config(num_workers):
+    yield ParallelConfig(backend="joblib", n_cpus_local=num_workers, wait_timeout=0.1)
```
AnesBenmerzoug, line 128: So we're only testing with joblib now? I was planning to do this to improve the tests' runtime but I was also planning to at least test one method using all the backends. Now we may have with bugs that we won't catch.

mdbenito: There is a whole file for testing all backends. If anything, the tests there should be improved, but it makes no sense to repeat the extremely costly computations three times (and at some point possibly more), I think.

src/pydvl/utils/parallel/futures/ray.py
mdbenito, line 356:
```suggestion
```

mdbenito, line 356:
```suggestion
```

AnesBenmerzoug (approved):

PR #375, opened 2023-07-12
Better inversion method info and remove concatenation from general

### LINKED ISSUE #374: Optimize IF calculations when using large datasets and models
I was working on the 4th experiment in the paper "Influence Functions in Deep Learning are Fragile", and I needed to train a ResNet-50 model on ImageNet. 

In this case (experiments with large models, large datasets), issues with **cuda memory error** and **redundant IF computations** may be magnified. I'd like to make some small suggestions to get pydvl better used in such scenarios:

1. In `compute_influence_factors()`, `test_grads = cat(test_grads)`. I got **cuda OOM errors** when giving a list of tensors with large dimensions. For example, when the shape of each individual tensor is `(batch_size=64, num_param_ResNet50=25557032)`. Reason behind is that `torch.cat()` makes a copy of tensors before concatenation, so there exist two identical tensors at the same time. 
One solution: we might allocate a tensor with a target dimension first and append results of `model.split_grad()` by slicing that tensor.
```
idx_start = 0
test_grads = torch.zeros(batch_size, num_param)
for x_test, y_test in maybe_progress(test_data, progress, desc="Batch Test Gradients"):
     bs = len(y_test)
     test_grads[idx_start:idx_start+bs] = model.split_grad(x_test, y_test, progress=False)
     idx_start += bs
```

2. Can we have an optional output, i.e. **Hessian matrix**, from function `solve_linear()`? Very often we are interested in also the Hessian and its eigenvalues in our experiments. So far, I have to call `solve_linear()` and then calculate the Hessian again separately, which is a bit redundant. 

3. Can we return also the **gamma value** in `solve_cg()`? That's for better determining the stopping criterion of CG and thus speeding up the inversion.

Xuzzo:
### Description

This PR closes #374 

### Changes

- Optimises memory usage in IF calculation
- Adds info about inversion of the hessian and forwards it to some of the main methods

### Checklist

- [x] Updated Changelog


src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -67,7 +72,8 @@ def solve_linear(
     matrix = model.hessian(
         all_x, all_y, progress=progress
     ) + hessian_perturbation * identity_tensor(model.num_params, device=model.device)
-    return torch.linalg.solve(matrix, b.T).T
+    info = {"hessian": matrix}
+    return torch.linalg.solve(matrix, b.T).T, info
```
AnesBenmerzoug, line 76: I think it would be better to change the interface to return a single dict with all the values instead of a tuple of tensor and dict.

AnesBenmerzoug: Or a named tuple .

src/pydvl/influence/general.py
```diff
@@ -64,12 +63,17 @@ def compute_influence_factors(
     :returns: An array of size (N, D) containing the influence factors for each
         dimension (D) and test sample (N).
     """
-    test_grads = []
-    for x_test, y_test in maybe_progress(
-        test_data, progress, desc="Batch Test Gradients"
+    test_grads = zero_tensor(
+        shape=(len(test_data.dataset), model.num_params),
+        dtype=test_data.dataset[0][0].dtype,
+    )
```
murong-xu, line 69: I found there is no device specified, the test_grads tensor is created on CPU while test_data could be on cuda.

src/pydvl/influence/general.py
```diff
@@ -102,16 +106,21 @@ def compute_influences_up(
...
-        number of train points.
+    :returns: An array of size [NxM], where N is number of influence factors, M
+        number of input points.
     """
-    train_grads = []
-    for x, y in maybe_progress(
-        input_data, progress, desc="Batch Split Input Gradients"
+    grads = zero_tensor(
+        shape=(len(input_data.dataset), model.num_params),
+        dtype=input_data.dataset[0][0].dtype,
```
murong-xu, line 114: Here is also the same problem, no device specified :)

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -67,7 +72,8 @@ def solve_linear(
     matrix = model.hessian(
         all_x, all_y, progress=progress
     ) + hessian_perturbation * identity_tensor(model.num_params, device=model.device)
-    return torch.linalg.solve(matrix, b.T).T
+    info = {"hessian": matrix}
```
murong-xu, line 75: Not sure here but are we interested in the original Hessian or the regularized one? In Fragile paper they computed eigenvalues on the original one.

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -111,10 +119,14 @@ def solve_batch_cg(
         total_grad_xy / total_points, v, backprop_on
     ) + hessian_perturbation * v.type(torch.float64)
     batch_cg = torch.zeros_like(b)
```
murong-xu, line 121: encountered cuda OOM error

src/pydvl/influence/general.py
murong-xu: Also: cuda OOM error occurred here

src/pydvl/influence/general.py
murong-xu: Also: cuda OOM error here

src/pydvl/influence/general.py
```diff
@@ -135,14 +144,20 @@ def compute_influences_pert(
...
+        the number of input data, and P the number of features.
     """
-    all_pert_influences = []
-    for x, y in maybe_progress(
-        input_data,
-        progress,
-        desc="Batch Influence Perturbation",
+    input_x = input_data.dataset[0][0]
+    all_pert_influences = zero_tensor(
+        shape=(len(input_data.dataset), len(influence_factors), *input_x.shape),
```
murong-xu, line 152: Cuda OOM might happen here

murong-xu (commented):
Thanks so much for your quick fix! I added some comments in the code, and I found the cuda OOM problem is still there (pointed out some places, pls see my comments).

Actually the tensor allocation cannot solve the memory error problem. Before I was using a small test set so it worked, but when I try to call `compute_influence_factors()` using all training data from ImageNet, the dimension of `test_grads` is `(1281167, 25557032)`, which requires 121976 GB of memory (our single GPU has 31.75GB). And decreasing the precision from float32 to float16 is still far not enough. 

I'm thinking about a few work arounds to solve it:
* batch-processing: without allocation of `test_grads`, call `solve_hvp()` directly after calculating `test_grad` for one test point `(x_test, y_test)` at a time. So the variable `b` will have a much smaller shape. But this is only suitable for CG method since in CG `test_grads` are iterated one by one.

* Save `test_grads` tensors in disk and create a Dataset class, s.t. we can generate batch of data with each `__getitem__` call. 
And I found one way os using memory mapped tensors to optimize the process: see https://discuss.pytorch.org/t/memory-mapped-tensor/8954/4 and 
https://towardsdatascience.com/how-to-boost-pytorch-dataset-using-memory-mapped-files-6893bff27b99
It describes how data in local storage can be optimized during file reading, and be mapped into memory, so that we can map files into memory that are much larger than the physical capacity of machine.

murong-xu:
Btw I tested the time it takes for CG to run. For ResNet-50 (25 mio parameters) on our single GPU, when I relaxed the criteria to `maxiter=250`, `rtol=1e-5`, `atol=1e-7`, it took 40s for finishing one data point. It's apparently stopped too early thus not the optimal solution, but considering running on the whole ImageNet...

src/pydvl/influence/frameworks/torch_differentiable.py
Xuzzo: done

src/pydvl/influence/frameworks/torch_differentiable.py
Xuzzo: should be done

src/pydvl/influence/frameworks/twice_differentiable.py
```diff
@@ -1,9 +1,11 @@
 from abc import ABC
+from collections import namedtuple
 from typing import Callable, Generic, List, Sequence, Tuple, TypeVar
 
 TensorType = TypeVar("TensorType", bound=Sequence)
 ModelType = TypeVar("ModelType")
 DeviceType = TypeVar("DeviceType")
+iHVPResult = namedtuple("iHVPResult", ["x", "info"])
```
AnesBenmerzoug, line 8: I think it's better to use the `NamedTuple` class from the `typing` package because it allows the use of type hints for the fields.

```suggestion
class iHVPResult(NamedTuple):
   x: TensorType
   info: Dict[str, Union[Any, TensorType]]
```

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -261,6 +283,19 @@ def cat(a: Sequence[torch.Tensor], **kwargs) -> torch.Tensor:
     return torch.cat(a, **kwargs)
 
 
+def zero_tensor(
+    shape: Sequence[int], dtype: Union[np.dtype, torch.dtype], **kwargs
+) -> torch.Tensor:
+    """Returns a tensor of shape :attr:`shape` filled with zeros."""
+    if isinstance(dtype, np.dtype):
+        dtype = TORCH_DTYPES[str(dtype)]
```
AnesBenmerzoug, line 291: Perhaps this would be more general and work better:

```suggestion
        dtype = getattr(torch, dtype.name)
```

I found it in [this issue](https://github.com/pytorch/pytorch/issues/40568#issuecomment-650198299)

src/pydvl/influence/inversion.py
```diff
@@ -58,7 +58,8 @@ def solve_hvp(
     :param progress: If True, display progress bars.
 
     :return: An array that solves the inverse problem,
```
AnesBenmerzoug, line 60:
```suggestion
    :return: An object that contains an array that solves the inverse problem,
```

AnesBenmerzoug:
@Xuzzo CI failure is due to #381. I will fix it now in a separate PR.

AnesBenmerzoug (approved):

murong-xu (approved):

schroedk (changes requested):
Please mark the change as breaking

PR #376, opened 2023-07-12
Improve IF docs and cleanup

### LINKED ISSUE #362: Remove dataset loading / generation from pydvl.utils.dataset
There are several functions in this module, like `synthetic_classification_dataset`, which do not belong in a library. They are only used in the notebooks and should be moved to the support library for them.

Xuzzo:
### Description

This PR closes #362 #374 

### Changes

- Optimises memory usage in IF calculation
- Adds info about inversion of the hessian and forwards it to some of the main methods
- Documentation for influence functions
- removes functions for loading datasets from the main library

### Checklist

- [x] Updated Changelog


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- Optimises memory usage in IF calculation and returns infor about hessian
```
schroedk, line 5: As the return type of a public function of the package is modified in a non backward compatible way, please mark this as a breaking change

src/pydvl/influence/frameworks/twice_differentiable.py
```diff
@@ -1,11 +1,28 @@
...
+    TypeVar,
+)
 
 TensorType = TypeVar("TensorType", bound=Sequence)
 ModelType = TypeVar("ModelType")
 DeviceType = TypeVar("DeviceType")
 
 
+@dataclass
+class iHVPResult(Generic[TensorType]):
```
schroedk, line 21: What is the reason to use a dataclass, instead of a tuple x, info ? In the end it is a question of taste, as we are breaking the function signature anyway:)

schroedk (changes requested):
Please mark the breaking change in CHANGELOG.md

schroedk (dismissed):

schroedk:
@Xuzzo there are tests, which randomly fail, due to stochastic parts in the code. (lissa, model training). I will fix this in #389

Xuzzo:
After discussion, we are keeping the dataclass for IhvpResult. I just added a way to easily unpack it and freeze the class after creation. Hence, I am closing #390

src/pydvl/influence/frameworks/twice_differentiable.py
```diff
@@ -1,11 +1,21 @@
...
+from dataclasses import dataclass
+from typing import Any, Callable, Dict, Generic, List, Sequence, Tuple, TypeVar
 
 TensorType = TypeVar("TensorType", bound=Sequence)
 ModelType = TypeVar("ModelType")
 DeviceType = TypeVar("DeviceType")
 
 
+@dataclass(frozen=True)
+class IhvpResult(Generic[TensorType]):
```
schroedk, line 11: Starting a class name with a capital I is in other languages marking an interface. What do you think about InverseHvpResult?

schroedk (changes requested):

schroedk (approved):

PR #378, opened 2023-07-14
Feature/344 functorch

schroedk:
### Changes

- modify the hessian computation to use torch.func

### Checklist

- [x] Updated Changelog
~~- [ ] If notebooks were added/changed, added boilerplate cells are tagged with `"nbsphinx":"hidden"`~~

### DIFF
--- a/requirements-notebooks.txt
+++ b/requirements-notebooks.txt
@@ -1,4 +1,4 @@
-torch==1.13.1
+torch==2.0.1
 torchvision==0.14.1
 datasets==2.6.1
 pillow==9.3.0
--- a/requirements-notebooks.txt
+++ b/requirements-notebooks.txt
@@ -1,4 +1,4 @@
 torch==2.0.1
-torchvision==0.14.1
+torchvision==0.15.2
 datasets==2.6.1
 pillow==9.3.0
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ setup(
     ],
     setup_requires=["wheel"],
     tests_require=["pytest"],
-    extras_require={"influence": ["torch"]},
+    extras_require={"influence": ["torch>=2.0.0"], "cupy": ["cupy-cuda11x>=12.1.0"]},
     author="appliedAI Institute gGmbH",
     long_description=long_description,
     long_description_content_type="text/markdown",
--- a/src/pydvl/influence/frameworks/functional.py
+++ b/src/pydvl/influence/frameworks/functional.py
@@ -5,15 +5,15 @@ import torch
 from torch.func import functional_call, grad, jvp, vjp
 from torch.utils.data import DataLoader
 
-from .util import Input_type, align_structure, to_model_device
+from .util import TensorContainerType, align_structure, to_model_device
 
 
 def hvp(
-    func: Callable[[Input_type], torch.Tensor],
-    params: Input_type,
-    vec: Input_type,
+    func: Callable[[TensorContainerType], TensorContainerType],
+    params: TensorContainerType,
+    vec: TensorContainerType,
     reverse_only: bool = True,
-) -> Input_type:
+) -> TensorContainerType:
     """
     Computes the Hessian-vector product (HVP) for a given function at given parameters.
     This function can operate in two modes, either reverse-mode autodiff only or both
@@ -37,6 +37,7 @@ def hvp(
         >>> hvp_vec = hvp(f, u, v)
         >>> assert torch.allclose(hvp_vec, torch.full((10, ), 2.0))
     """
+    output: TensorContainerType
 
     if reverse_only:
         _, vjp_fn = vjp(grad(func), params)
@@ -71,16 +72,14 @@ def batch_hvp_gen(
 
     for inputs, targets in iter(data_loader):
         batch_loss = batch_loss_function(model, loss, inputs, targets)
-        yield partial(
-            hvp, batch_loss, dict(model.named_parameters()), reverse_only=reverse_only
-        )
+        yield partial(hvp, batch_loss, dict(model.named_parameters()), reverse_only=reverse_only)  # type: ignore
 
 
 def empirical_loss_function(
     model: torch.nn.Module,
     loss: Callable[[torch.Tensor, torch.Tensor], torch.Tensor],
     data_loader: DataLoader,
-) -> Callable[[Dict[str, torch.Tensor]], torch.Tensor]:
+) -> Callable[[Dict[str, torch.Tensor]], Dict[str, torch.Tensor]]:
     """
     Creates a function to compute the empirical loss of a given model on a given dataset.
     If we denote the model parameters with $\theta$, the resulting function approximates
--- a/src/pydvl/influence/frameworks/functional.py
+++ b/src/pydvl/influence/frameworks/functional.py
@@ -8,6 +8,11 @@ from torch.utils.data import DataLoader
 from .util import TensorContainerType, align_structure, to_model_device
 
 
+__all__ = [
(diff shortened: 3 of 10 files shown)


requirements-notebooks.txt
```diff
@@ -1,4 +1,4 @@
-torch==1.13.1
+torch==2.0.1
```
AnesBenmerzoug, line 1: Could you also add the minimum version to the `extras_require` field inside setup.py?

src/pydvl/influence/frameworks/functional.py
```diff
@@ -0,0 +1,201 @@
+from functools import partial
+from typing import Callable, Dict, Generator, Iterable
+
+import torch
+from torch.func import functional_call, grad, jvp, vjp
+from torch.utils.data import DataLoader
+
+from .util import Input_type, align_structure, to_model_device
+
```
AnesBenmerzoug, line 9: Could you please add an `__all__` variable to the top of module to better document which functions are expected to be used outside the module?

src/pydvl/influence/frameworks/functional.py
```diff
@@ -0,0 +1,201 @@
...
+def batch_loss_function(
+    model: torch.nn.Module,
+    loss: Callable[[torch.Tensor, torch.Tensor], torch.Tensor],
+    x: torch.Tensor,
+    y: torch.Tensor,
+) -> Callable[[Dict[str, torch.Tensor]], torch.Tensor]:
+    """
+    Creates a function to compute the loss of a given model on a given batch of data.
+
+    Args:
```
AnesBenmerzoug, line 129: This should use the rst docstring style.

src/pydvl/influence/frameworks/functional.py
```diff
@@ -0,0 +1,201 @@
...
+) -> Callable[[Dict[str, torch.Tensor]], torch.Tensor]:
+    """
+    Creates a function to compute the empirical loss of a given model on a given dataset.
+    If we denote the model parameters with $\theta$, the resulting function approximates
+
+    .. math::
+
+        f(\theta) = \frac{1}{N}\sum_{i=1}^N \operatorname{loss}(y_i, \operatorname{model}(\theta, x_i)))
+
+    Args:
```
AnesBenmerzoug, line 92: This should use the rst docstring style like thie `get_hvp_function` function below.

src/pydvl/influence/frameworks/functional.py
```diff
@@ -0,0 +1,201 @@
...
+        model (torch.nn.Module): The model for which the loss should be computed.
+        loss (Callable[[torch.Tensor, torch.Tensor], torch.Tensor]): The loss function to be used.
+        data_loader (torch.utils.data.DataLoader): The data loader for iterating over the dataset.
+
+    Returns:
+        Callable[[torch.nn.parameter.Parameter], torch.Tensor]: A function that computes the empirical loss
+                                                               of the model on the dataset for given model parameters.
```
AnesBenmerzoug, line 99:
```suggestion
    :param model:The model for which the loss should be computed.
    :param loss: The loss function to be used.
    :param data_loader: The data loader for iterating over the dataset.

    :return: A function that computes the empirical loss
                  of the model on the dataset for given model parameters.
```

src/pydvl/influence/frameworks/functional.py
```diff
@@ -0,0 +1,201 @@
...
+
+    Returns:
+        Input_type: The HVP of the function at the given parameters with the given vector.
+
+    Examples:
+        >>> def f(z): return torch.sum(z**2)
+        >>> u = torch.ones(10, requires_grad=True)
+        >>> v = torch.ones(10)
+        >>> hvp_vec = hvp(f, u, v)
+        >>> assert torch.allclose(hvp_vec, torch.full((10, ), 2.0))
```
AnesBenmerzoug, line 38:
```suggestion
   :param func: The function for which the HVP is computed.
   :param params: The parameters at which the HVP is computed.
   :param vec: The vector with which the Hessian is multiplied.
   :param reverse_only: Whether to use only reverse-mode autodiff
            (True, default) or both forward- and reverse-mode autodiff (False).

    :return: Input_type: The HVP of the function at the given parameters with the given vector.

    :Example:
       
     >>> def f(z): return torch.sum(z**2)
     >>> u = torch.ones(10, requires_grad=True)
     >>> v = torch.ones(10)
     >>> hvp_vec = hvp(f, u, v)
     >>> assert torch.allclose(hvp_vec, torch.full((10, ), 2.0))
```

src/pydvl/influence/frameworks/functional.py
```diff
@@ -0,0 +1,201 @@
...
+
+    Args:
+        model (torch.nn.Module): The model for which the loss should be computed.
+        loss (Callable[[torch.Tensor, torch.Tensor], torch.Tensor]): The loss function to be used.
+        x (torch.Tensor): The input data for the batch.
+        y (torch.Tensor): The true labels for the batch.
+
+    Returns:
+        Callable[[Dict[str, torch.Tensor]], torch.Tensor]: A function that computes the loss
+                                                           of the model on the batch for given model parameters.
```
AnesBenmerzoug, line 137:
```suggestion
   :param model: The model for which the loss should be computed.
   :param loss: The loss function to be used.
   :param x: The input data for the batch.
   :param y: The true labels for the batch.

   :return: A function that computes the loss
                  of the model on the batch for given model parameters.
```

AnesBenmerzoug (changes requested):
The docstrings should use the same rst style.

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -505,3 +524,119 @@ def solve_arnoldi(
...
+            raise ValueError(
+                "Without setting an explicit device, cupy is not supported"
+            )
+
+        def mv_cupy(x):
+            x = from_dlpack(x.toDlpack())
+            y = hessian_vp(x) + hessian_perturbation * x
+            return cp.from_dlpack(to_dlpack(y))
+
+        eigen_vals, eigen_vecs = eigsh(
```
AnesBenmerzoug, line 591: There is a lot of repetition and return points in this function. I think it should be simplified by making a single call to `eigsh` and defining the different variables and functions in the if and else branches, along with an extra function to convert the result to torch tensors.

setup.py
```diff
@@ -21,7 +21,7 @@
     ],
     setup_requires=["wheel"],
     tests_require=["pytest"],
-    extras_require={"influence": ["torch"]},
+    extras_require={"influence": ["torch>=2.0.0"], "cupy": ["cupy-cuda11x>=12.1.0"]},
```
AnesBenmerzoug, line 24: cupy is only used with influence functions, right? What would be the disadvantage of adding it to the influence extras instead of creating a separate one just for it?

AnesBenmerzoug:
@schroedk could you please merge develop into your branch and mark this PR as ready so that CI can run?

src/pydvl/influence/frameworks/torch_differentiable.py
schroedk: Agree, done

setup.py
schroedk: On mac, there is no wheel cupy-cuda11x. If we include it into the influence extra, then no mac user could install the influence extra

AnesBenmerzoug: Alright, we can leave it for now.
I will create a separate issue to consider using [environment markers](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#environment-markers) for this.

AnesBenmerzoug (approved):

PR #383, opened 2023-07-26
Decouple ray.init from ParallelConfig

### LINKED ISSUE #373: Coupled ParallelConfig and ray.init()
The code https://github.com/appliedAI-Initiative/pyDVL/blob/839a22c83a36b6e19fd0a5c7f3eb68472662accf/src/pydvl/utils/parallel/futures/ray.py#L68-L69 couples `ParallelConfig` to the parameters of ray.init(): adding new properties to `ParallelConfig` raises a `RuntimeError`, e.g.:
```
RuntimeError: Unknown keyword argument(s): wait_timeout
```

AnesBenmerzoug:
### Description

This PR closes #373 

### Changes

- Explicitly pick values from ParallelConfig when initializing a parallel backend.

### Checklist

- [x] Updated Changelog

### DIFF
--- a/src/pydvl/utils/parallel/backend.py
+++ b/src/pydvl/utils/parallel/backend.py
@@ -106,8 +106,8 @@ class JoblibParallelBackend(BaseParallelBackend, backend_name="joblib"):
 
     def __init__(self, config: ParallelConfig):
         self.config = {
-            "n_jobs": config.n_cpus_local,
             "logging_level": config.logging_level,
+            "n_jobs": config.n_cpus_local,
         }
 
     def get(
@@ -158,11 +158,9 @@ class RayParallelBackend(BaseParallelBackend, backend_name="ray"):
     """
 
     def __init__(self, config: ParallelConfig):
-        self.config = {"logging_level": config.logging_level}
-        if config.address is None:
+        self.config = {"address": config.address, "logging_level": config.logging_level}
+        if self.config["address"] is None:
             self.config["num_cpus"] = config.n_cpus_local
-        else:
-            self.config["address"] = config.address
         if not ray.is_initialized():
             ray.init(**self.config)
         # Register ray joblib backend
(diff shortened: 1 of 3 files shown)


src/pydvl/utils/parallel/backend.py
schroedk, line 1: This might be a bit picky. I was wondering why there is a full config in the constructor, although the specific implementations only need some parts. Then I realized, there are some necessary checks in the init of ParallelConfig, so all good. But it led me to the following problem. The BaseParallelBackend is not enforcing, that you can call it's child classes to be called with an instance of ParallelConfig, i.e. having an init with config being the only parameter. The fix would be to add an abstract class method _from_config(config: ParallelConfig), which every child has to implement. Then the init can very and the construction logic is encapsuled in the _from_config (you would call _create in this function). Then in the init_parallel_backend, you would call _from_config instead of _create. This would blow up the interface a little bit, but on the other hand cleans the init of the implementations and makes sure that every new backend implementation follows the interface. As I said very minor and picky. What do you think?

src/pydvl/utils/parallel/backend.py
```diff
@@ -162,12 +158,11 @@ class RayParallelBackend(BaseParallelBackend, backend_name="ray"):
     """
 
     def __init__(self, config: ParallelConfig):
-        config_dict = asdict(config)
-        config_dict.pop("backend")
-        n_cpus_local = config_dict.pop("n_cpus_local")
-        if config_dict.get("address", None) is None:
-            config_dict["num_cpus"] = n_cpus_local
-        self.config = config_dict
+        self.config = {"logging_level": config.logging_level}
```
kosmitive, line 161: The issue was created due to the modification of the temp dir. In that case where would we add it, as a separate parameter to __init__ or to ParallelConfig?

src/pydvl/utils/parallel/backend.py
AnesBenmerzoug: Thanks for the suggestion Kristof. Unfortunately I don't think I understood it. Could you please explain it in more detail in an issue? 

To be honest, I personally think we should get rid of the `ParallelConfig` class: we can tell the users to call `ray.init` if they want to connect to a ray cluster or configure how many CPUs a local one would use and we can tell them to configure joblib just like it is done in scikit-learn. The only thing we have to do ourselves would be to make something similar for the Futures interface.

src/pydvl/utils/parallel/backend.py
AnesBenmerzoug: Is this comment for the [class-wise shapley PR](https://github.com/appliedAI-Initiative/pyDVL/pull/338)?

src/pydvl/utils/parallel/backend.py
schroedk: See #385

schroedk:
@AnesBenmerzoug there are tests, which randomly fail, due to non-deterministic model init and training. I will fix this in another PR.

schroedk (approved):

PR #389, opened 2023-08-01
Refactor tests_influences:

### LINKED ISSUE #388: Testing torch related modules is non-deterministic
Tests for the influence module, in particular those testing the lissa and arnoldi inversion method randomly fail with AssertionErrors,
due to the difference not being in the desired tolerance. Due to the stochastic nature of lissa and training routines needed for arnoldi tests, we should make those test deterministic.

schroedk:
### Description

This PR closes #388

### Changes

* use dataclasses for test case definition
* make test deterministic

### Checklist

- ~~[ ] Updated Documentation (if necessary)~~
- ~~[ ] Updated Changelog~~

### DIFF
--- a/tests/influence/test_influences.py
+++ b/tests/influence/test_influences.py
@@ -1,4 +1,5 @@
-from typing import Dict, Tuple
+from dataclasses import dataclass
+from typing import Callable, Dict, Tuple
 
 import numpy as np
 import pytest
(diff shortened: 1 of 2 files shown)


tests/influence/test_influences.py
```diff
@@ -77,6 +78,13 @@ def analytical_linear_influences(
     return result
 
 
+@pytest.fixture(scope="function")
+def deterministic_seed():
```
AnesBenmerzoug, line 82: We already have [2 fixtures](https://github.com/appliedAI-Initiative/pyDVL/blob/develop/tests/conftest.py#L61-L76) that are used to seed random numbe generators.
Perhaps they can be combined (at least `pytorch_seed` and `deterministic_seed`?

tests/influence/test_influences.py
```diff
@@ -413,3 +476,15 @@ def test_influences_arnoldi(
     )
 
     assert np.allclose(direct_influence, low_rank_influence, rtol=1e-1)
+
+
+@pytest.mark.torch
+@pytest.mark.parametrize("run", [1, 2])
+def test_deterministic_seed(deterministic_seed, run):
```
AnesBenmerzoug, line 483: is this test really needed? It tests PyTorch's behaviour which is something that we do not control anyway.

AnesBenmerzoug:
In case you want to test an algorithm/method that is fast enough to run and without pinning the seed of the random number generators, we have a special [`tolerate`](https://github.com/appliedAI-Initiative/pyDVL/blob/develop/tests/conftest.py#L374) fixture for that than can be used to run a particular call multiple times and allow a maximum number of failures.

tests/influence/test_influences.py
schroedk: Yeah, agree, will remove it

tests/influence/test_influences.py
schroedk: We could add torch.use_deterministic_algorithms to pytorch_seed. This should work then. What do you think?

AnesBenmerzoug (approved):

PR #393, opened 2023-08-07
setting device to lanzcos_low_rank_hessian_approx

### LINKED ISSUE #391: Device error in solve_arnoldi()
When calling `solve_arnoldi()`, I got `RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument mat2 in method wrapper_CUDA_mm)` in the code section below:
```
result = low_rank_representation.projections @ (
        torch.diag_embed(1.0 / low_rank_representation.eigen_vals)
        @ (low_rank_representation.projections.t() @ b.t())
    )
```
where the tensors `projections` and `eigen_vals` are on cpu but `b` is on cuda.

Although it seems to already pass an argument `device` to the `lanzcos_low_rank_hessian_approx()`, this device information is not used when computing eigenvalues on CPU (by default `eigen_computation_on_gpu=False`).

Xuzzo:
### Description

This PR closes #391 

### Changes

- adds a warning for low rank approximation when device is gpu
- re allocates eigenvals and eigenvecs to gpu before returning in lanzcos_low_rank_hessian_approx

### Checklist

- [ ] Updated Changelog


schroedk (commented):
@Xuzzo I will test the changes via running tests on GPU

src/pydvl/influence/frameworks/torch_differentiable.py
```diff
@@ -668,7 +676,7 @@ def mv(x):
 
         eigen_vals, eigen_vecs = e.eigenvalues, e.eigenvectors
 
-    eigen_vals = to_torch_conversion_function(eigen_vals)
-    eigen_vecs = to_torch_conversion_function(eigen_vecs)
+    eigen_vals = to_torch_conversion_function(eigen_vals).to(device)
```
schroedk, line 679: After a little digestion phase, I am not sure, if this is the correct place to add the .to(device).
The reason is the following, we have three potential computation scenarios:

1. All on CPU
2. All on GPU
3. Mixed, model forward pass on GPU -> Arnoldi iteration on CPU

Case 1 and 2 are clear, but the third case can be a valid use case, when the projections are too large to fit on GPU memory. If we force the array back on the GPU here, it could be a problem.

Maybe handling this in downstream code, i.e. solve_arnoldi, and improve documentation? What do you think?

schroedk:
@Xuzzo Most probably, there will be conflicts with #394 . I will resolve them, depending on which branch is merged first.

schroedk (approved):

Issue #343, opened 2023-04-18
Nyström's method for approximate inverse of the Hessian


mdbenito:
[1] uses a low rank approximation and some algebra to invert the Hessian.

Basically:
1.Select k << p random indices of H \in R^{p x p}
2. Build a submatrix H_k with k columns, then a square submatrix H_kk of that one with k rows, and compute its pseudoinverse H_kk^~1
3. Apply some matrix arithmetics to compute an inverse of a slight perturbation of H_k @ H_kk^~1 @ H_k^T

In practice, one does not build H and select from it. There is some code [here](https://github.com/moskomule/hypergrad/blob/main/hypergrad/approx_hypergrad.py#L105) that looks pretty solid. It boils down to < 20 lines. Worth a try?

**[1]** _Hataya, Ryuichiro, and Makoto Yamada. “Nystrom Method for Accurate and Scalable Implicit Differentiation.” In 26th International Conference on Artificial Intelligence and Statistics (AISTATS), 2023. https://doi.org/10.48550/arXiv.2302.09726._

PR #394, opened 2023-08-07
Refactoring/influence

schroedk:
### Description

This PR introduces a cleanup on how we can handle different frameworks. I would like to start from this, to add the jax framework and an additional method (Nyström approximation #343). This is not breaking the user interface again, as it is only modifying internal mechanisms of the package.

### Changes

- make the top level modules independent of torch, not just by name
- introduce an inversion method registry to easily add new methods

### Checklist

- [x] Updated Changelog


src/pydvl/influence/general.py
```diff
@@ -63,25 +59,30 @@ def compute_influence_factors(
...
+    tensor_util: Type[TensorUtilities] = TensorUtilities.from_twice_differentiable(
+        model
     )
-    for batch_idx, (x_test, y_test) in enumerate(
-        maybe_progress(test_data, progress, desc="Batch Test Gradients")
+    stack = tensor_util.stack
+    unsqueeze = tensor_util.unsqueeze
+    cat = tensor_util.cat
+
+    test_grads = []
```
Xuzzo, line 69: appending tensors to a list and then stacking is typically slower and uses more memory than just filling a previously create one. Any reason to go back to the old setup in all these methods?

Xuzzo (commented):
Nice new design! Thanks a lot. I just have one comment on the concatenation of grads.

src/pydvl/influence/general.py
schroedk: @Xuzzo I would like to postpone this optimization question, as abstracting away the bracket access is a further complexity.

schroedk: @Xuzzo I would like to postpone this optimization question, as abstracting away the bracket access is a further complexity.

schroedk: @Xuzzo I would like to postpone this optimization question, as abstracting away the bracket access is a further complexity.

src/pydvl/influence/general.py
```diff
@@ -56,32 +48,36 @@ def compute_influence_factors(
     :param model: A model wrapped in the TwiceDifferentiable interface.
     :param training_data: A DataLoader containing the training data.
     :param test_data: A DataLoader containing the test data.
-    :param inversion_func: function to use to invert the product of hvp (hessian
-        vector product) and the gradient of the loss (s_test in the paper).
+    :param inversion_method: name of method for computing inverse hessian vector products
```
mdbenito, line 51:
```suggestion
    :param inversion_method: name of method for computing inverse hessian vector
        products.
```

src/pydvl/influence/inversion.py
```diff
@@ -53,51 +52,135 @@ def solve_hvp(
...
+        :raises TypeError: If the provided model_type or inversion_method are of the wrong type.
+        :raises ValueError: If the function to be registered does not match the required signature.
+        :return: A decorator for registering a function.
+        """
+
+        if not isinstance(model_type, type):
+            raise TypeError(f"'model_type' must be a Type[TwiceDifferentiable]")
+
+        if not isinstance(inversion_method, InversionMethod):
+            raise TypeError(f"'inversion_method' must be an InversionMethod")
```
mdbenito, line 109:
```suggestion
        if not isinstance(inversion_method, InversionMethod):
            raise TypeError(f"'inversion_method' must be an 'InversionMethod' "
                            f"but has type {type(inversion_method)} instead.")
```

src/pydvl/influence/inversion.py
```diff
@@ -53,51 +52,135 @@ def solve_hvp(
...
+            def wrapper(*args, **kwargs):
+                return func(*args, **kwargs)
+
+            cls.registry[key] = wrapper
+            return wrapper
+
+        return decorator
+
+    @classmethod
+    def get_registered(
```
mdbenito, line 148: The suffix "registered" seems redundant. Consider its use: `InversionRegistry.get_registered(...)` vs `InversionRegistry.get(...)`. As a rule of thumb, I think we should prefer conciseness if it makes code read out aloud in a natural way.

```suggestion
    def get(
```

src/pydvl/influence/inversion.py
```diff
@@ -53,51 +52,135 @@ def solve_hvp(
...
+        [TwiceDifferentiable, DataLoaderType, TensorType, float], InverseHvpResult
+    ]:
+        key = (model_type, inversion_method)
+        method = cls.registry.get(key, None)
+        if method is None:
+            raise ValueError(f"No function registered for {key}")
+        return method
+
+    @classmethod
+    def call_registered(
```
mdbenito, line 160: Same as above (and for consistency too), I would replace this with:
```suggestion
    def call(
```

src/pydvl/influence/inversion.py
```diff
@@ -53,51 +52,135 @@ def solve_hvp(
...
+        :param b: Array as the right hand side of the equation $Ax = b$.
+        :param hessian_perturbation: Regularization of the hessian.
+        :param kwargs: Additional keyword arguments to pass to the inversion method.
+
+        :return: An object that contains an array that solves the inverse problem,
+            i.e. it returns $x$ such that $Ax = b$, and a dictionary containing
+            information about the inversion process.
+        """
+
+        return cls.get_registered(type(model), inversion_method)(
```
mdbenito, line 184:
```suggestion
        return cls.get(type(model), inversion_method)(
```

src/pydvl/influence/torch/functional.py
```diff
@@ -75,14 +77,26 @@ def batch_hvp_gen(
 
     for inputs, targets in iter(data_loader):
         batch_loss = batch_loss_function(model, loss, inputs, targets)
-        yield partial(hvp, batch_loss, dict(model.named_parameters()), reverse_only=reverse_only)  # type: ignore
+
+        def batch_hvp(vec: torch.Tensor):
+            model_params = dict(model.named_parameters())
```
mdbenito, line 82: I would move this outside the closure, same as `batch_loss`

src/pydvl/influence/torch/functional.py
```diff
@@ -180,27 +194,25 @@ def get_hvp_function(
         k: p if track_gradients else p.detach() for k, p in model.named_parameters()
     }
 
-    def hvp_function(vec: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor]:
+    def hvp_function(vec: torch.Tensor) -> torch.Tensor:
         v = align_structure(params, vec)
         empirical_loss = empirical_loss_function(model, loss, data_loader)
-        return hvp(empirical_loss, params, v, reverse_only=reverse_only)
+        return flatten_tensors_to_vector(
```
mdbenito, line 200: Sorry for the naive question: This is probably not the right place nor the only one to make this comment, but there seem to be a few places where lists of tensors are constructed and then concatenated, which would double memory requirements. In this case, given an input tensor and the list of parameters, it should be possible to preallocate the output tensor for _all_ the jvp/vjp operations, right? Maybe the slicing would become a bit tricky, though. I'm not sure where the biggest memory issues arise, but is this case on the radar? Or is it not worth considering?

src/pydvl/influence/torch/torch_differentiable.py
```diff
@@ -19,42 +18,393 @@
...
+        grad_f = torch.autograd.grad(
+            loss_value, self.parameters, create_graph=create_graph
+        )
+        return flatten_tensors_to_vector(grad_f)
+
+    def hessian(self, x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
+        """Calculates the explicit hessian of model parameters given data ($x$ and $y$).
+        :param x: A matrix [NxD] representing the features $x_i$.
+        :param y: A matrix [NxK] representing the target values $y_i$.
+        :returns: A tensor representing the hessian of the model, i.e. the second derivative wrt. the model parameters.
```
mdbenito, line 106:
```suggestion
        :returns: A tensor representing the hessian of the loss wrt. the model parameters.
```

mdbenito (changes requested):
I did a somewhat superficial review since we already discussed most design choices. Left some ball-busting nit-picking comments, but I think this is great step forward.

mdbenito (approved):

src/pydvl/influence/general.py
schroedk: @Xuzzo pre-allocating result tensor is implemented

src/pydvl/influence/torch/functional.py
schroedk: Created an issue #401

Issue #392, opened 2023-08-05
Explicit randomization for subprocesses


kosmitive:
As sub-processes fork the random generator from their parent process, it might happen that different sub processes obtain the same sequence of random numbers in different processes. Depending on the backend it might be an issue. In general it would be advantageous to make it explicit, e.g.

```python
def set_time_pid_seed():
    t = int(os.getpid() + time.time())
    np.random.seed(t)
    random.seed(t)
```

or even better, to pass different seeds from the parent process to the forked child processes. The latter would give us reproducability, see also (https://joblib.readthedocs.io/en/latest/auto_examples/parallel_random_state.html)

AnesBenmerzoug:
This should be related to #242

PR #396, opened 2023-08-08
Add solution for explicit randomization in subprocesses.

### LINKED ISSUE #242: Properly handle random seeds
In order to make the results completely reproducible, we should be able to pass random seeds to the valuation methods and have all subsequent random number generation seeded from that.

This was inspired by some changes in PR #238

kosmitive:
### Description

This PR closes #242, #392 and #398

### Changes

- Added seed to functions from `pydvl.utils.numeric`, `pydvl.value.shapley` and
  `pydvl.value.semivalues`. 
- Added new tests for testing reproducibility of modified functions.
- Introduced new type `Seed` and conversion function `ensure_seed_sequence`.

### Postponed

- Randomization for PyTorch dependent algorithms and modules #397.
- Full reproducibility due to random order of execution #416.


AnesBenmerzoug:
This isn't the recommended way to implement seeded random number generation. We should favour creating `rng` objects (in the case of numpy) and a [`Random`](https://docs.python.org/3.11/library/random.html#random.Random) object (in the case of pure Python starting from version 3.9) and use those for any subsequent generation.

Here's how it's done in scikit-learn: https://github.com/scikit-learn/scikit-learn/blob/2a2772a87b6c772dc3b8292bcffb990ce27515a8/sklearn/utils/validation.py#L1207

We could use that as inspiration, even though it's using the legacy [`RandomState`](https://numpy.org/doc/stable/reference/random/legacy.html#numpy.random.RandomState) class instead of the new [`Generator`](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.Generator) class.

This doesn't really close #242. To solve that we should expose a `seed` parameter in all algorithms that we can use to seed the different workers.

Refer to [this](https://joblib.readthedocs.io/en/latest/auto_examples/parallel_random_state.html) and [this](https://numpy.org/doc/stable/reference/random/parallel.html) for more details.

kosmitive:
@AnesBenmerzoug Can we drop support for Python 3.8. Reason for that is that in `numpy>=1.25`, the Generator object has a spawn method. This method doesn't exist in previous versions. However Python 3.8 doesn't requires `numpy<1.25`. Not having this method makes the design for seeding more complex as one has to distinguish between `SeedSequence` and `Generator`.

src/pydvl/utils/numeric.py
```diff
@@ -59,21 +60,29 @@ def num_samples_permutation_hoeffding(eps: float, delta: float, u_range: float)
     return int(np.ceil(np.log(2 / delta) * 2 * u_range**2 / eps**2))
 
 
-def random_subset(s: NDArray[T], q: float = 0.5) -> NDArray[T]:
+def random_subset(
+    s: NDArray[T],
+    q: float = 0.5,
+    seed: numbers.Integral | np.random.SeedSequence | np.random.Generator = None,
```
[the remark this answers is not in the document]
mdbenito: > This syntax is only supported starting from 3.10. You should use `Union` instead

Note that we use `from future import __annotations__` in several files to support this operator.

src/pydvl/utils/parallel/map_reduce.py
```diff
@@ -1,6 +1,10 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
mdbenito: This provides support for the `|` operator to declare union types, among other things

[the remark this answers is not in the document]
AnesBenmerzoug: That's for the information. I didn't know that.

src/pydvl/utils/numeric.py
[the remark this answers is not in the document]
AnesBenmerzoug: I see. I think that for the sake of consistency we should continue the types from the typing module and leave this change for another issue.

kosmitive:
> @kosmitive
> 
>     1. No, we shouldn't drop Python 3.8 until it reaches its end-of-life (14 Oct 2024). The `Generator`'s `spawn` is just a convenience that was added in numpy 1.25 and we shouldn't bump the minimum requirement just for that. We can just directly instantiate a [SeedSequence](https://numpy.org/doc/stable/reference/random/bit_generators/generated/numpy.random.SeedSequence.html#numpy.random.SeedSequence) object and call its `spawn` method.
> 
>     2. Yes, please do that so we can be sure that this change works. No need to run all tests twice though to check that seeding works.
> 
>     3. Yes, that makes sense. Anyway, the random seed handling is different in PyTorch.

1. Okay then I allow only to pass a SeedSequence prior to using parallel execution.

3. I'll create an issue for further investigation.

src/pydvl/utils/numeric.py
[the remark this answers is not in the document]
mdbenito: > I see. I think that for the sake of consistency we should continue the types from the typing module and leave this change for another issue.

Since there are several files using it, it's already too late. We can systematically add it at some point, but I don't think it's a big issue.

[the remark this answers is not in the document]
kosmitive: Already reverted it.

src/pydvl/utils/parallel/map_reduce.py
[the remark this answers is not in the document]
mdbenito: @AnesBenmerzoug To be precise, the import does not really provide support, it just disables evaluation of annotations, so the interpreter doesn't choke on unsupported types for the operator, but the type checker still can do its job.

mdbenito (commented):
Thanks for the PR. I have left a couple of comments, but I think we need to discuss the general design for this. It is important to distinguish the different concepts used and how they enter the picture at which stages.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 pyDeprecate>=0.3.2
-numpy>=1.20
+numpy>=1.24
```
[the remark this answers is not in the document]
AnesBenmerzoug: No, it isn't.

src/pydvl/utils/types.py
```diff
@@ -64,3 +74,23 @@ def wrapper(*args, **kwargs):
         return fun(*args, **kwargs)
 
     return wrapper
+
+
+Seed = Optional[Union[int, np.random.SeedSequence]]
+SeedOrGenerator = Union[Seed, np.random.Generator]
```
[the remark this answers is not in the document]
kosmitive: Problem is `numpy<1.25`. In upper versions Generator and SeedSequence share the same spawn method. If we support Python 3.8 we can't update numpy and this need to stick to that pattern, where two cases appear:

1. Functions which parallelize need a SeedSequence. There is no beautiful way to go from Generator back to SeedSequence. 
2. Functions which aren't parallelized, can also take a Generator instance. If we want to reduce it we have to stick here with SeedSequence, because case 1 can't be set to Generator.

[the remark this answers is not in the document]
kosmitive: @mdbenito

kosmitive:
@mdbenito @AnesBenmerzoug Let's decide on one way. Do we want to go forward with a seed parameter approach or do we switch to a different approach. What are your thoughts?

src/pydvl/value/result.py
```diff
@@ -657,7 +658,11 @@ def to_dataframe(
 
     @classmethod
     def from_random(
-        cls, size: int, total: Optional[float] = None, **kwargs
+        cls,
+        size: int,
+        total: Optional[float] = None,
+        seed: SeedOrGenerator = None,
```
AnesBenmerzoug, line 664:
```suggestion
        seed: Optional[SeedOrGenerator] = None,
```

src/pydvl/value/sampler.py
```diff
@@ -90,6 +101,7 @@ def __init__(
         indices: NDArray[T],
         index_iteration: IndexIteration = IndexIteration.Sequential,
         outer_indices: NDArray[T] = None,
+        seed: SeedOrGenerator = None,
```
AnesBenmerzoug, line 104:
```suggestion
        seed: Optional[SeedOrGenerator] = None,
```

src/pydvl/value/semivalues.py
```diff
@@ -99,8 +104,11 @@ def _semivalues(
     :param done: Stopping criterion.
     :param progress: Whether to display progress bars for each job.
     :param job_id: id to use for reporting progress.
+    :param seed: Either an instance of a numpy random number generator or a seed for it.
     :return: Object with the results.
     """
+    log.warning(f"Seed is {seed}")
```
AnesBenmerzoug, line 110: This shouldn't be needed anymore

```suggestion
```

src/pydvl/utils/types.py
```diff
@@ -64,3 +74,19 @@ def wrapper(*args, **kwargs):
...
+def ensure_seedsequence(seed: Seed) -> Optional[SeedSequence]:
+    """Check if the seed is valid and return a SeedSequence object if it is. If it is
+    not valid, return None."""
+
+    if seed is None:
+        return None
+    elif isinstance(seed, int):
+        return SeedSequence(seed)
+    else:
+        return seed
```
AnesBenmerzoug, line 92: This could be the implementation for what I described in my other comment.

```suggestion
def ensure_seedsequence(seed: Optional[int, Generator, SeedSequence] = None) -> SeedSequence:
    """If the passed seed is a SeedSequence object then it is returned as is.
    Otherwise, a new object is created."""
    if isinstance(seed, SeedSequence):
        return seed
    elif isinstance(seed, Generator):
        return seed.bit_generator._seed_seq
    else:
        return SeedSequence(seed)
```

src/pydvl/utils/types.py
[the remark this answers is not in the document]
kosmitive: That is a good idea. I took over the later proposed function ensure_seed_seq below.

src/pydvl/utils/types.py
kosmitive: See https://github.com/appliedAI-Initiative/pyDVL/pull/396#discussion_r1309621511

src/pydvl/utils/types.py
```diff
@@ -92,3 +95,26 @@ def __call__(cls, *args, **kwargs):
...
+    Otherwise, a new SeedSequence object is created from the passed (optional) seed.
+
+    :param seed: Either an int, a Generator object a SeedSequence object or None.
+
+    :returns: A SeedSequence object.
+    """
+    if isinstance(seed, SeedSequence):
+        return seed
+    elif isinstance(seed, Generator):
+        return seed._bit_generator._seed_seq  # noqa
```
AnesBenmerzoug, line 118: One fewer private member access:

```suggestion
        return seed.bit_generator._seed_seq  # noqa
```

https://numpy.org/doc/1.20/reference/random/generated/numpy.random.Generator.bit_generator.html#numpy.random.Generator.bit_generator

src/pydvl/utils/numeric.py
```diff
@@ -227,3 +249,7 @@ def top_k_value_accuracy(
     top_k_pred_values = np.argsort(y_pred)[-k:]
     top_k_accuracy = len(np.intersect1d(top_k_exact_values, top_k_pred_values)) / k
     return top_k_accuracy
+
+
+def sets_are_equal(subset_1: NDArray[np.int_], subset_2: NDArray[np.int_]):
+    return list(subset_1) == list(subset_2)
```
AnesBenmerzoug, line 255: This should be better than a conversion:

```suggestion
    return np.isin(subset_1, subset_2).all()
```

Could you please add a docstring to this function?

src/pydvl/value/sampler.py
```diff
@@ -142,17 +144,10 @@ def complement(self, exclude: Sequence[T]) -> NDArray[T]:
...
-        """Iterates over indices in the order specified at construction.
-
-        FIXME: this is probably not very useful, but I couldn't decide
-          which method is better
-        """
+        """Iterates over indices in the order specified at construction."""
         if self._index_iteration is PowersetSampler.IndexIteration.Sequential:
             for idx in self._outer_indices:
                 yield idx
-        elif self._index_iteration is PowersetSampler.IndexIteration.Random:
```
AnesBenmerzoug, line 153: You just removed the else case but it still exists and is not handled.

src/pydvl/value/sampler.py
```diff
@@ -205,6 +200,24 @@ def weight(cls, n: int, subset_len: int) -> float:
         ...
 
 
+class StochasticSampler(PowersetSampler[T], abc.ABC):
```
AnesBenmerzoug, line 203: You just set `index_iteration`

src/pydvl/value/semivalues.py
```diff
@@ -99,7 +100,7 @@ def __call__(self, n: int, k: int) -> float:
 
 def _marginal(u: Utility, coefficient: SVCoefficient, sample: SampleT) -> MarginalT:
     """Computation of marginal utility. This is a helper function for
-    :func:`semivalues`.
+    :func:`semivalues`. TODO Allow
```
AnesBenmerzoug, line 103: TODO?

src/pydvl/utils/parallel/check.py
```diff
@@ -0,0 +1,47 @@
+import inspect
+from functools import partial
+from typing import Callable
+
+
+def check_fn_accepts_parameter(fn: Callable, param_name: str) -> bool:
```
AnesBenmerzoug, line 6: There is already [a function](https://github.com/appliedAI-Initiative/pyDVL/blob/4f65403278b4f77956bf7c0ef50250f5a5b4dcf2/src/pydvl/utils/types.py#L47) that does something similar.

src/pydvl/value/sampler.py
mdbenito, line 203: I think this is clear case for a Mixin, to avoid an additional level in the class hierarchy.

src/pydvl/value/semivalues.py
```diff
@@ -252,14 +254,17 @@ def compute_shapley_semivalues(
...
         address, number of cpus, etc.
     :param progress: Whether to display a progress bar.
 
     :return: Object with the results.
     """
+    sampler_instance = sampler_t(u.data.indices)
+    sampler_instance.seed = seed
     return semivalues(
-        sampler_t(u.data.indices),
+        sampler_instance,
```
mdbenito, line 267: If you used a mixin (and added it first in the inheritance list) then this could be:

```python
return semivalues(
        sampler_t(u.data.indices, seed=seed),
```

Note that you have to be careful about the resolution order and passing along arguments. Here's a PoC:

```python
class A:
    def __init__(self, a, **kwargs):
        super().__init__(**kwargs)  # Pass along any unused arguments
        self.a = a

class M:
    def __init__(self, m, **kwargs):
        super().__init__(**kwargs)  # Pass along any unused arguments
        self.m = m

class B(M, A):
    def __init__(self, a, b, **kwargs):
        super().__init__(a=a, **kwargs)
        self.b = b

b = B(a=1, b=2, m=7)

print(b.a)  # 1
print(b.b)  # 2
print(b.m)  # 7
```

src/pydvl/utils/numeric.py
mdbenito: The conversion to list would fail to implement set equality because the order would matter. Nevertheless I think the function is too trivial to be added. Also, conversion to `set` is probably more intuitive

```python 
set(a) == set(b)
```

So I would just delete this and do the conversion inline wherever necessary

src/pydvl/value/sampler.py
kosmitive: Remove inheritance from `PowersetSampler[T]`.

src/pydvl/value/sampler.py
kosmitive: Readded.

src/pydvl/value/sampler.py
```diff
@@ -257,7 +265,7 @@ def weight(cls, n: int, subset_len: int) -> float:
         return float(2 ** (n - 1)) if n > 0 else 1.0
 
 
-class UniformSampler(StochasticSampler[T]):
+class UniformSampler(PowersetSampler[T], StochasticSamplerMixin):
```
mdbenito, line 268: @kosmitive Are you sure you don't want to swap the order of inheritance here? Recall that python resolves methods in the order you use here. If you want to be able to initialise UniformSampler with a seed, which would be an argument for the mixin, you need to ensure that the `__init__` of the mixin is called first. See the example [in my other comment](https://github.com/appliedAI-Initiative/pyDVL/pull/396#discussion_r1310585383)

src/pydvl/utils/numeric.py
kosmitive: @AnesBenmerzoug Removed function and thus docstring not necessary, for equality one would need both inclusions `np.setxor1d(a, b).size == 0`.

src/pydvl/utils/parallel/check.py
kosmitive: Merged both functions.

src/pydvl/value/semivalues.py
kosmitive: Done

src/pydvl/value/sampler.py
kosmitive: Found a different solution, where it doesn't depend on the order.

mdbenito (commented):
I think this is almost ready

src/pydvl/value/sampler.py
```diff
@@ -395,3 +434,8 @@ def __iter__(self) -> Iterator[SampleT]:
     @classmethod
     def weight(cls, n: int, subset_len: int) -> float:
         return float(2 ** (n - 1)) if n > 0 else 1.0
+
+
+StochasticSampler = Union[
```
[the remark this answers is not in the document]
kosmitive: `mypy` throws an error when doing this. It needs an explicit cast to go from `PermutationSampler` to `StochasticSampler`, see

```bash
src/pydvl/value/semivalues.py:243: error: Incompatible default for argument "sampler_t" (default has type "Type[PermutationSampler[Any]]", argument has type "Type[StochasticSampler[Any]]")
src/pydvl/value/semivalues.py:283: error: Incompatible default for argument "sampler_t" (default has type "Type[PermutationSampler[Any]]", argument has type "Type[StochasticSampler[Any]]")
src/pydvl/value/semivalues.py:323: error: Incompatible default for argument "sampler_t" (default has type "Type[PermutationSampler[Any]]", argument has type "Type[StochasticSampler[Any]]")
src/pydvl/value/semivalues.py:376: error: Incompatible default for argument "sampler_t" (default has type "Type[PermutationSampler[Any]]", argument has type "Type[StochasticSampler[Any]]")
```

mdbenito:
@kosmitive Merging the documentation branch into develop has caused a lot of conflicts, sorry.

src/pydvl/value/sampler.py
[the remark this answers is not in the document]
mdbenito: Hmm you are right, my bad. I guess those would have to be protocols... what a pain. Then it will have to remain a union. It's just too ugly. Also, on top of the problems mentioned, one has to remember to update the type after adding new samplers. And any workaround with dynamic type inference, if possible, will be a bloated mess... :/

src/pydvl/utils/parallel/backlog.py
```diff
@@ -0,0 +1,37 @@
...
+from typing import Callable, Generator, Generic, List, Tuple, TypeVar
+
+T = TypeVar("T")
+
+
+def wrap_index(*args, i: int, fn: Callable, **kwargs):
+    return i, fn(*args, **kwargs)
+
+
+class Backlog(Generic[T]):
```
mdbenito, line 12: Please don't add this functionality to this PR. I would have a dozen comments to make and we don't have time if we want to include the seeding in 0.7.0. Revert this commit and make a separate PR

kosmitive: But it isn't reproducible without. So in that case we have to move the test cases for reproducibility as well.

src/pydvl/value/sampler.py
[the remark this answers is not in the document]
kosmitive: I hope that Github gives a notification once the linked issue is closed.

src/pydvl/utils/parallel/backlog.py
kosmitive: Removed and put into https://github.com/aai-institute/pyDVL/pull/417 for further discussion.

mdbenito:
@kosmitive [This error](https://github.com/aai-institute/pyDVL/actions/runs/6050840796/job/16421062158?pr=396#step:5:1399) indicates that you must have made some mistake merging develop into your branch, because the function `semivalues` was removed from that branch and its tests.

src/pydvl/utils/functional.py
```diff
@@ -0,0 +1,39 @@
...
+import inspect
+from functools import partial
+from typing import Callable, Set, Tuple
+
+__all__ = ["unroll_partial_fn_args"]
+
+
+def unroll_partial_fn_args(fun: Callable) -> Set[str]:
+    """
+    Unroll a function that was set by functools.partial.
```
mdbenito, line 12: What's "unrolling"? From the docstring it's unclear what this does. Please explain and add an example

src/pydvl/utils/numeric.py
```diff
@@ -111,24 +122,32 @@ def random_powerset(
         ValueError: if the element sampling probability is not in [0,1]
 
     """
+    if not isinstance(s, np.ndarray):
```
mdbenito, line 125: We don't do runtime type checking anywhere. Why here?

src/pydvl/utils/types.py
```diff
@@ -42,6 +46,27 @@ def score(self, x: NDArray, y: NDArray) -> float:
         pass
 
 
+def call_fun_remove_arg(*args, fun: Callable, arg: str, **kwargs):
```
mdbenito, line 49: What is gained by having this function? Why not leave it in `maybe_remove_arg` since it's only used there?

src/pydvl/utils/types.py
```diff
@@ -42,6 +46,27 @@ def score(self, x: NDArray, y: NDArray) -> float:
...
+    """
+    Calls the given function with the given arguments, but removes the given argument.
```
mdbenito, line 51: Cryptic docstring.

src/pydvl/utils/types.py
```diff
@@ -58,18 +83,10 @@ def maybe_add_argument(fun: Callable, new_arg: str) -> Callable:
     Returns:
         A new function accepting one more keyword argument.
     """
-    params = inspect.signature(fun).parameters
-    if new_arg in params.keys():
+    if new_arg in unroll_partial_fn_args(fun):
```
AnesBenmerzoug, line 86: What is the logic behind this change?

src/pydvl/utils/types.py
```diff
@@ -95,3 +112,47 @@ def __call__(cls, *args, **kwargs):
...
+    Args:
+        fn: The function to execute.
+        args: The arguments to pass to the function.
+        seeds: The seeds to use.
+        kwargs: The keyword arguments to pass to the function.
+
+    Returns:
+        A tuple of the results of the function.
+    """
+    return tuple(fn(*deepcopy(args), **deepcopy(kwargs), seed=seed) for seed in seeds)
```
mdbenito, line 158: I'm a bit wary of this hidden deepcopy. Datasets and Utilities are heavy objects. Would a user of this function expect the deep copy to happen?

src/pydvl/value/shapley/gt.py
```diff
@@ -194,6 +202,8 @@ def group_testing_shapley(
         config: Object configuring parallel computation, with cluster
             address, number of cpus, etc.
         progress: Whether to display progress bars for each job.
+        seed: Either an instance of a numpy random number generator or a seed
+        for it.
```
mdbenito, line 206:
```suggestion
        seed: Either an instance of a numpy random number generator or a seed
            for it.
```

src/pydvl/value/shapley/gt.py
```diff
@@ -134,11 +140,12 @@ def _group_testing_shapley(
         n_samples: total number of samples (subsets) to use.
         progress: Whether to display progress bars for each job.
         job_id: id to use for reporting progress (e.g. to place progres bars)
-
+        seed: Either an instance of a numpy random number generator or a seed
+        for it.
```
mdbenito, line 144:
```suggestion
        seed: Either an instance of a numpy random number generator or a seed
            for it.
```

src/pydvl/value/shapley/common.py
```diff
@@ -85,6 +89,8 @@ def compute_shapley_values(
             criteria using boolean operators. Some methods ignore this argument,
             others require specific subtypes.
         n_jobs: Number of parallel jobs (available only to some methods)
+        seed: Either an instance of a numpy random number generator or a seed
+        for it.
```
mdbenito, line 93:
```suggestion
        seed: Either an instance of a numpy random number generator or a seed
            for it.
```

src/pydvl/value/result.py
```diff
@@ -690,7 +695,8 @@ def from_random(
                 ("efficiency" property of Shapley values).
             kwargs: Additional options to pass to the constructor of
                 [ValuationResult][pydvl.value.result.ValuationResult]. Use to override status, names, etc.
-
+            seed: ither an instance of a numpy random number generator or a seed for
+            it.
```
mdbenito, line 699:
```suggestion
            seed: Either an instance of a numpy random number generator or a seed for
                it.
```

src/pydvl/value/result.py
```diff
@@ -690,7 +695,8 @@ def from_random(
                 ("efficiency" property of Shapley values).
             kwargs: Additional options to pass to the constructor of
                 [ValuationResult][pydvl.value.result.ValuationResult]. Use to override status, names, etc.
```
mdbenito, line 697:
```suggestion
            kwargs: Additional options to pass to the constructor of
                [ValuationResult][pydvl.value.result.ValuationResult]. Use to
                override status, names, etc.
```

src/pydvl/utils/types.py
```diff
@@ -95,3 +112,47 @@ def __call__(cls, *args, **kwargs):
...
+    """
+    if isinstance(seed, SeedSequence):
+        return seed
+    elif isinstance(seed, Generator):
+        return cast(SeedSequence, seed.bit_generator.seed_seq)  # type: ignore
+    else:
+        return SeedSequence(seed)
+
+
+def call_fn_multiple_seeds(
```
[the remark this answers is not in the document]
mdbenito: Agreed

mdbenito (commented):
More comments. Please double check all docstrings before requesting another review.

src/pydvl/utils/types.py
kosmitive: Once a partial function is passed to `maybe_add_argument`, the inner ``__call__`` logic is neglected for these types and thus also if it is contained in the arguments. Furthermore `partial` has to be used, because otherwise the wrapped function can't be pickled (happens for `seed` but not for `job_id`). So recursive `partial` support is needed.

src/pydvl/utils/types.py
kosmitive: It is required to be pickable. Once the function is wrapped (happens for `seed` and not for `job_id`) it can not be pickled anymore by the default pickle library. To not over-complicate the code by introducing `dill` or `cloudpickle`, I moved the function to the outer-scope.

src/pydvl/utils/types.py
kosmitive: It is mostly for tests, moved to tests module.

src/pydvl/utils/types.py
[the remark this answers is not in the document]
kosmitive: Moved to `test/value/utils.py`

src/pydvl/utils/functional.py
kosmitive: Added docstring.

src/pydvl/utils/numeric.py
kosmitive: Removed.

src/pydvl/utils/types.py
kosmitive: Reformulated.

src/pydvl/utils/types.py
kosmitive: I wouldn't expect it to happen, but it would annoy me if it wasn't there :D

src/pydvl/utils/types.py
mdbenito: Note that ray depends on cloudpickle, it's pretty much standard. But it's good to avoid dependencies. Then please document the choice in the code, so as to avoid these discussions. Please **always** document decisions which might not be obvious to other readers of the code, because we waste a lot of time otherwise.

kosmitive: Done

mdbenito (approved):
@kosmitive I went ahead and pushed directly to your branch because I don't know how to explain better that you need to think of the readers when writing documentation. I also moved things around a bit and made a couple minor changes / simplifications.

PR #403, opened 2023-08-24
Improve documentation of influence package:

schroedk:
* add more information
* change to google style format

### Description

Google style docstrings for influence package

### Checklist

- ~~[ ] Wrote Unit tests (if necessary)~~
- ~~[ ] Updated Changelog~~


src/pydvl/influence/__init__.py
```diff
@@ -1,10 +1,8 @@
 """
 This package contains algorithms for the computation of the influence function.
 
-.. warning::
-   Much of the code in this package is experimental or untested and is subject
-   to modification. In particular, the package structure and basic API will
-   probably change.
+> **Warning:** Much of the code in this package is experimental or untested and is subject to modification.
```
AnesBenmerzoug, line 4: Do we still need this warning even after the refactoring?

schroedk: I would like to keep it for now.

schroedk:
Added reference to pytorch objects

schroedk:
> @schroedk mkdocstrings has support for referencing other project's documentation using [inventories](https://mkdocstrings.github.io/usage/?h=inventory#cross-references-to-other-projects-inventories). Here's the for PyTorch: https://pytorch.org/docs/stable/objects.inv. I think we should add that.
> 
> Did you manage to find a fix for the references inside docstrings?

Added reference to pytorch objects.
It was only possible to insert the references directly into the python files.

Issue #367, opened 2023-06-27
Ray via pip not working for Apple ARM-based CPU


schroedk:
It seems, that in order to use ray with ARM-based Apple CPUs (M1, M2), one has to install a certain dependency of ray via conda package manager

`ImportError while loading conftest '/Users/kristof/Projects/pyDVL/tests/conftest.py'.
tests/conftest.py:9: in <module>
    import ray
.tox/base/lib/python3.10/site-packages/ray/__init__.py:101: in <module>
    _configure_system()
.tox/base/lib/python3.10/site-packages/ray/__init__.py:65: in _configure_system
    raise ImportError(
E   ImportError: Failed to import grpc on Apple Silicon. On Apple Silicon machines, try `pip uninstall grpcio; conda install grpcio`. Check out https://docs.ray.io/en/master/ray-overview/installation.html#m1-mac-apple-silicon-support for more details.
ERROR: InvocationError for command /Users/kristof/Projects/pyDVL/.tox/base/bin/pytest tests/utils -vv --cov .tox/base/lib/python3.10/site-packages/pydvl -m 'not torch' (exited with code 4)`

Installing via conda works (tested on a M2 machine). Should we handle this somehow or put a note in the README?

schroedk:
@mdbenito

mdbenito:
See #380 about that.

mdbenito:
If we merge #408 this is no longer our concern

Issue #261, opened 2023-02-10
Ditch sphinx


mdbenito:
Let's move to mkdocs. Nothing can be worse than sphinx. See https://github.com/appliedAI-Initiative/pymetrius/issues/8

mdbenito:
See https://resources.gitpages.aai.sh/mlops/mike-mkdocs/docs/0.1.1/ for versioning of the docs

PR #408, opened 2023-08-26
Make ray an optional dependency

### LINKED ISSUE #380: Make ray dependency optional
Now that we have a joblib parallel backend and it is set as default, thanks to #355, we should consider making `ray` an optional dependency.

mdbenito:
### Description

This PR closes #380 #367 .

### Changes
Because we are already using a sub-class registry, there was little to do:

- Moved ParallelBackends into separate modules
- Guard imports
- Adapt the tests to skip if ray is not available. Not very elegant, but the alternative was to change the fixture `parallel_backend` and that would have meant a lot more changes. Ideas? (I'd just leave it as is tbh)
- **Documentation is missing**, waiting until we have #261 to avoid silly conflicts

### Checklist

- [x] Updated Changelog

### DIFF
--- a/src/pydvl/utils/parallel/backends/__init__.py
+++ b/src/pydvl/utils/parallel/backends/__init__.py
@@ -1,0 +1,9 @@
+try:
+    from .joblib import *
+except ImportError:
+    pass
+
+try:
+    from .ray import *
+except ImportError:
+    pass
--- a/src/pydvl/utils/parallel/backends/__init__.py
+++ b/src/pydvl/utils/parallel/backends/__init__.py
@@ -1,7 +1,4 @@
-try:
-    from .joblib import *
-except ImportError:
-    pass
+from .joblib import *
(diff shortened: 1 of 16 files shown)


src/pydvl/utils/parallel/backends/__init__.py
```diff
@@ -0,0 +1,9 @@
+try:
+    from .joblib import *
```
AnesBenmerzoug, line 2: Are there plans to make joblib optional? Since it's the default then we should always have it as a requirement and we shouldn't have this try except block.

AnesBenmerzoug (approved):

src/pydvl/utils/parallel/backends/__init__.py
mdbenito: That's me not thinking clearly. You are correct, of course.

PR #409, opened 2023-08-30
Add binder links to rendered notebooks

### LINKED ISSUE #405: Add binder links to mkdocs
These were introduced in 3c744892c3717eda709d017c07954d9b0a3de0ce for sphinx

AnesBenmerzoug:
### Description

This PR closes #405 

image

### Changes

- Uses `mknotebooks` instead of `mkdocs-jupyter`.
- Added hook to replace binder url created by mknotebooks.

### Checklist

- [ ] Updated Changelog

### DIFF
--- a/docs/css/extra.css
+++ b/docs/css/extra.css
@@ -111,11 +111,6 @@ body[data-md-color-scheme="slate"] .invertible img {
 }
 
 /* Rendered dataframe from jupyter */
-div.rendered_html {
-    width: 100% !important;
-    overflow-x: scroll;
-}
-
 table.dataframe {
     display: block;
     max-width: -moz-fit-content;
--- a/docs/css/extra.css
+++ b/docs/css/extra.css
@@ -109,3 +109,18 @@ body[data-md-color-scheme="default"] .invertible img {
 body[data-md-color-scheme="slate"] .invertible img {
     filter: invert(100%) hue-rotate(180deg);
 }
+
+/* Rendered dataframe from jupyter */
+div.rendered_html {
+    width: 100% !important;
+    overflow-x: scroll;
+}
+
+table.dataframe {
+    display: block;
+    max-width: -moz-fit-content;
+    max-width: fit-content;
+    margin: 0 auto;
+    overflow-x: auto;
+    white-space: nowrap;
+}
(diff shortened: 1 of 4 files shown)


docs/css/extra.css
```diff
@@ -109,3 +109,18 @@ body[data-md-color-scheme="default"] .invertible img {
 body[data-md-color-scheme="slate"] .invertible img {
     filter: invert(100%) hue-rotate(180deg);
 }
+
+/* Rendered dataframe from jupyter */
+div.rendered_html {
+    width: 100% !important;
+    overflow-x: scroll;
```
mdbenito, line 116: This creates scrollbars all over the place on my system:
image

Removing it seems not to cause any problem. Is it necessary for you?

mdbenito (commented):
In this branch I see tons of warnings from mkdocs that weren't there before, like the following:

```
WARNING -  Doc file '_build/html/examples/shapley_utility_learning.ipynb' contains a relative link '../pydvl/value/shapley/common.rst#pydvl.value.shapley.common.compute_shapley_values', but the target '_build/html/pydvl/value/shapley/common.rst' is not found among documentation files.
```
Any idea why?'

Also: the launch binder button is enough, the "binder" caption / subtitle is redundant and quite ugly. I would limit the whole thing to the anchor tag:

image

docs/css/extra.css
AnesBenmerzoug: Done.

AnesBenmerzoug:
@mdbenito Unlike [`mkdocs-jupyter`](https://github.com/danielfrg/mkdocs-jupyter/blob/main/mkdocs_jupyter/plugin.py#L107), [`mknotebooks`](https://github.com/greenape/mknotebooks/blob/master/mknotebooks/plugin.py#L263) converts the notebooks to markdown and lets mkdocs handle the conversion from markdown to HTML. That makes it work with other plugins and extensions. That's why those warnings started appearing. We could now convert all the links to the jupyter notebooks to use aliases and let mkdocs resolve them but that makes the links unreadable in the notebook itself.

AnesBenmerzoug:
I removed the Binder caption.

mdbenito:
> @mdbenito Unlike [`mkdocs-jupyter`](https://github.com/danielfrg/mkdocs-jupyter/blob/main/mkdocs_jupyter/plugin.py#L107), [`mknotebooks`](https://github.com/greenape/mknotebooks/blob/master/mknotebooks/plugin.py#L263) converts the notebooks to markdown and lets mkdocs handle the conversion from markdown to HTML. That makes it work with other plugins and extensions. That's why those warnings started appearing. We could now convert all the links to the jupyter notebooks to use aliases and let mkdocs resolve them but that makes the links unreadable in the notebook itself.

I see. But the links are already broken in the notebooks since they are relative to the built docs, right? It think we could convert them.

PR #412, opened 2023-08-30
Fix some types

mdbenito:
### Description

More type fixing. This fixes errors appearing with higher versions of mypy, but not [the two errors in the pipeline](https://github.com/appliedAI-Initiative/pyDVL/actions/runs/6028045241/job/16354504551) which I do not understand and don't want to just sidestep with `# type: ignore`

### Checklist

- [ ] Updated Changelog

### DIFF
--- a/src/pydvl/utils/parallel/futures/ray.py
+++ b/src/pydvl/utils/parallel/futures/ray.py
@@ -386,7 +386,7 @@ class _WorkItemManagerThread(threading.Thread):
                 # We cancel the future's object references
                 # We cannot cancel a running future object.
                 for future in self.submitted_futures:
-                    ray.cancel(future.object_ref)
+                    ray.cancel(future.object_ref)  # type: ignore
                 # Make sure we do this only once to not waste time looping
                 # on running processes over and over.
                 executor._cancel_futures &= ~CancellationPolicy.RUNNING
--- a/src/pydvl/value/stopping.py
+++ b/src/pydvl/value/stopping.py
@@ -31,6 +31,8 @@ truth tables of [Status][pydvl.utils.status.Status]. The unary operator `~`
 these operations affect the behavior of the stopping criteria.
 """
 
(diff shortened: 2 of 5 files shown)


src/pydvl/utils/parallel/futures/ray.py
```diff
@@ -386,7 +386,7 @@ def flag_executor_shutting_down(self):
                 # We cancel the future's object references
                 # We cannot cancel a running future object.
                 for future in self.submitted_futures:
-                    ray.cancel(future.object_ref)
+                    ray.cancel(future.object_ref)  # type: ignore
```
mdbenito, line 389: I couldn't find the type definition for ray...Future

src/pydvl/value/stopping.py
```diff
@@ -195,9 +197,9 @@ def __invert__(self) -> "StoppingCriterion":
 
 def make_criterion(
     fun: StoppingCriterionCallable,
-    converged: Callable[[], NDArray[np.bool_]] = None,
-    completion: Callable[[], float] = None,
-    name: str = None,
+    converged: Callable[[], NDArray[np.bool_]] | None = None,
+    completion: Callable[[], float] | None = None,
+    name: str | None = None,
```
AnesBenmerzoug, line 202: We're mixing style. We have both `Optional[...]` and `... | None`

src/pydvl/utils/parallel/futures/ray.py
AnesBenmerzoug: `Future` comes from `concurrent.futures` and `object_ref` is set [at runtime](https://github.com/appliedAI-Initiative/pyDVL/blob/54d9f40dd29412d0e1f0de589e759b3c5418c177/src/pydvl/utils/parallel/futures/ray.py#L238) so typing wouldn't work anyway.

AnesBenmerzoug (approved):

PR #422, opened 2023-09-04
Update publish.yaml

AnesBenmerzoug:
### Description

This PR adds missing env variables to the publish CI job
that prevented pushing new docs.

Hopefully with this fix we can trigger the deployment of a new PyPI package using the develop branch's CI file.

### Changes

- Defined GITHUB_BOT* env variable in publish workfloa.

### Checklist

- [ ] Updated Changelog


mdbenito:
Sorry, I’m lost here. Why is that env var necessary and what problem are we trying to solve here?

AnesBenmerzoug:
Those 2 env vars represent the Github Bot user. They are used when pushing a new version of the docs to the gh-pages branch and since they were missing it was always failing. I didn't notice it before because they were just passed as empty strings (default for undefined env vars).

Issue #291, opened 2023-02-28
Use codecov


mdbenito:
That. In pymetrius, see https://github.com/appliedAI-Initiative/pymetrius/issues/5

- [x] Create an account and link it to the org.
- [ ] Disable coverage in local tests
 
Any repo can [apparently configure this](https://codecovpro.zendesk.com/hc/en-us/articles/15821363818523-Who-is-my-Codecov-Admin-). Just [go here](https://app.codecov.io/gh/aai-institute/pyDVL/new)

PR #429, opened 2023-09-13
Reduce execution time of tests

### LINKED ISSUE #312: Reduce execution time of tests
- [ ] Further split tests into groups
- [ ] Remove redundant tests
- [ ] Decrease runtime of longest tests
- [ ] Ensure that we are truly using both CPUs in the runner

AnesBenmerzoug:
### Description

This PR closes #312 #291

### Changes

- Separate tests and notebook tests dispatch workflows.
- Add CI step to upload coverage report to codecov.
- Use `--failed-first` pytest flag to start testing from the last failed tests.
- Generate coverage report as xml file.
- Increase number of slowest test durations to show to 30.
- Use `pytest-split` to automatically split the tests in 4 groups based on their measured duration.
- Commit `.test_durations` file for use in CI.
- Use `pytest-cases` to separate the test case definition from the tests in the influences tests.
- Refactor influence tests to reduce their runtime. 
- Define new pytest marker `pytest.mark.slow` to mark slow tests and skip them by default.
- Define new custom pytest flag `--slow-tests` to enable running slow tests.
- Combine tox test environments into `tests` environment.
- Remove unused `report` tox environment.
- Update contributing docs.
- Pin mypy to version `1.5.1`.
- Fix type hints.
- Cache mypy cache directory on CI.
- Add `pytest-sugar` as a dev dependency.
- Define pytest cache class and fixture that uses cloudpickle to accelerate heavy fixtures.
- Update and fix some of the notebooks.
- Use `nbmake` to test notebooks instead of using custom script.
- Fix links to documentation in README.

### Checklist

- [x] Updated Changelog


AnesBenmerzoug:
@mdbenito do we have a codecov account?

mdbenito:
> @mdbenito do we have a codecov account?

Yes, it is active for this repo only, and [any admin can apparently configure it](https://codecovpro.zendesk.com/hc/en-us/articles/15821363818523-Who-is-my-Codecov-Admin-)

CONTRIBUTING.md
```diff
@@ -92,28 +92,60 @@ It is possible to pass optional command line arguments to pytest, for example to
...
+- `-n` sets the number of parallel workers for 
+  [pytest-xdist](https://pytest-xdist.readthedocs.io/en/latest/).
+  
+  There are two layers of parallelization in the tests.
+  An inner one within the tests themselves, i.e. the parallelism in the algorithms,
+  and an outer one by pytest-xdist. The latter is controlled by the `-n` argument.
+  If you experience segmentation faults with the tests,
+  try running them with `-n 0` to disable parallelization.
+
+- `--slow-tests` enables running slow tests.
```
mdbenito, line 121: I think this merits some explanation. What are "slow tests", why do we have such a category, when are they run by default (they always run on CI, correct?), etc.

CONTRIBUTING.md
```diff
@@ -92,28 +92,60 @@ It is possible to pass optional command line arguments to pytest, for example to
...
+groups of them of separately. Use `pytest --markers` to get a list and description
+of all available markers.
+
+Two important markers are:
+
+- `pytest.mark.slow` which is used to mark slow tests and skip them by default.
+
+  You can run additionaly run slow tests using: `pytest --slow-tests`
+
+- `pytest.mark.torch` which is used to mark tests that require PyTorch.
```
mdbenito, line 135: Why do we treat pytorch separately? If that's because it's an optional dependency, so is ray, and so will be memcached. So maybe we need other marks for those? Or maybe we don't need this one for pytorch. If the module is installed, we run the tests, always, like ray, etc.

CONTRIBUTING.md
```diff
@@ -301,6 +333,20 @@ sizeable amount of time, so care must be taken not to overdo it:
 1. All algorithm tests must be on very simple datasets and as quick as possible
 2. We try not to trigger CI pipelines when unnecessary (see [Skipping CI
 runs](#skipping-ci-runs)).
+3. We split the tests based on their duration into groups and run them in parallel.
+  
+   For that we use [pytest-split](https://jerry-git.github.io/pytest-split)
+   to first store the duration of all tests with `pytest --store-durations pytest --slow-tests`
+   in a `.test_durations` file.
+   
+   Then we can have as many splits as we want:
```
mdbenito, line 342: How are the commands below to be interpreted? I suppose each one is to be run in a different machine, because the splits into groups are deterministic, correct?

src/pydvl/influence/torch/functional.py
mdbenito, line 1: Where do all these changes for influence functions come from?

src/pydvl/value/oob/oob.py
```diff
@@ -30,8 +27,8 @@ def compute_data_oob(
     *,
     n_est: int = 10,
     max_samples: float = 0.8,
-    loss: Callable = None,
-    n_jobs: int = None,
+    n_jobs: Optional[int] = None,
+    loss: Optional[LossFunction] = None,
```
mdbenito, line 31: why this permutation? n_jobs is usually among the last parameters.

src/pydvl/value/semivalues.py
```diff
@@ -272,7 +272,7 @@ def compute_generic_semivalues(
                     # Filter out samples for indices that have already converged
                     filtered_samples = samples
                     if skip_converged and len(done.converged) > 0:
-                        # cloudpickle can't pickle this on python 3.8:
+                        # TODO: cloudpickle can't pickle this on python 3.8:
```
mdbenito, line 275: I wouldn't say this is a to-do. It was just a comment explaining why I removed the nicer filter with a lambda

tests/conftest.py
```diff
@@ -54,6 +50,22 @@ def pytest_addoption(parser):
     )
 
 
+@pytest.fixture
```
mdbenito, line 53: * Caching tests is kind of risky. This is meant only for fixtures correct?
* How many of those do we have and how much time is saved with the cache? I only see the linear shapley values. Are there more?
* If there were more, wouldn't it be better to have a `@pytest.cached_fixture` decorator that acted as a memoizator? I.e. it hashes the inputs, maybe the pip versions, or whatever and so on.

src/pydvl/value/semivalues.py
AnesBenmerzoug: I just added the TODO to make it easier to find once we stop supporting 3.8 which should happen next year.

src/pydvl/influence/torch/functional.py
AnesBenmerzoug: Type checks were failing and I had to make a few changes to make both type checks and tests run.

CONTRIBUTING.md
AnesBenmerzoug: I created an issue (#453) for this.

CONTRIBUTING.md
AnesBenmerzoug: I updated CONTRIBUTING.md with more detail.

Things aren't really faster with xdist because of the configuration we are using. For me xdist only starts a single worker so it doesn't change anything.

CONTRIBUTING.md
AnesBenmerzoug: I updated CONTRIBUTING.md with more detail.

Yes, each command is supposed to run in a separate job and yes the splits should be deterministic.

src/pydvl/value/oob/oob.py
AnesBenmerzoug: Must have been accidental. I will permute them back.

tests/conftest.py
AnesBenmerzoug: - Yes, it is only for test fixtures. Actually there is already such a feature in pytest but it uses json for serialization so I couldn't really use it.
- We only use it for one fixture for now and it divides the test runtime by half.
- I didn't want to optimize it too much because I wasn't sure how it would actually be used and whether we would keep it.

src/pydvl/influence/torch/functional.py
mdbenito: Should've guess myself, thanks.

tests/conftest.py
mdbenito: By half! Holy 🐮! Ok, maybe then clearly document that it's only meant for fixtures, and if possible add some warning / error / whatever? To deter people from using it on actual tests, or to make them (us) aware of the intended purpose.

CONTRIBUTING.md
mdbenito: Maybe you don't have that many cores? It should instantiate (n_cpus-1) / 4 workers. The 4 comes from the inner parallelization of the algorithms, which we may want to reduce to 3 or even 2 if that allows starting more xdist workers. I have 32 cores so I am happily running very quick tests 😇

AnesBenmerzoug: Yes, I saw that and understood why that was happening. We can probably tune the values later if it turns out we need another speedup.

tests/conftest.py
AnesBenmerzoug: The main use of pytest's cache is to allow using the `--failed-first` flag to start a test run with the last failed tests.

It says in the documentation that other extension and plugins may access the cache but the fact that it only supports json serialization probably means that it shouldn't be used the way that we're using it.

Still, since for now it is actually saving us some time and it is only used in one single location we can keep it.

See the documentation for more information: https://docs.pytest.org/en/7.1.x/how-to/cache.html

mdbenito (dismissed):
This is awesome!

I'd just wish that `pytest-split` could replace `xdist` entirely, but there are two issues:

* Crucially: Is pytest-split supported by pycharm? 
* We would need some automated way of starting the maximum number of split jobs possible given the amount of CPUs. Or maybe it'd be enough to just have a config variable (env?) setting this quantity.

tests/conftest.py
mdbenito: Can you document this where it is used? It's going to be hard to remember this in half a year

CONTRIBUTING.md
mdbenito: Hmm... let me see if I we see things the same way:
* xdist is used locally, max number of cores, divided by 4. Probably could pick less than 4 processes for inner parallelization
* split is used on CI only, or locally, manually

AnesBenmerzoug:
@mdbenito 

- What do you mean by "supported by pycharm"?
- Do you mean starting multiple groups in parallel? pytest-split doesn't support that. The best we can do is to either increase or decrease the number of groups.

mdbenito:
> * What do you mean by "supported by pycharm"?

I mean, if I run tox as a run configuration, will I get IDE support for the tests, in particular with links to the right files (as opposed to useless copies in the tox environment)

> * Do you mean starting multiple groups in parallel? pytest-split doesn't support that. The best we can do is to either increase or decrease the number of groups.

I mean running the n commands in one go, and computing the number n automatically.

But never mind. xdist manages to fully use the machine and split works wonders on CI, so let's leave it like that.

mdbenito:
@AnesBenmerzoug One more thing: I think it would make sense to deactivate coverage reports in local tests. It's pointless, generates tons of temp files and surely slows things down.

mdbenito:
See #454 in case you want to look into it for this PR, otherwise we can merge this as-is.

AnesBenmerzoug:
@mdbenito 

- I will document the custom cache fixture as well as when to rerun pytest with `--store-durations`.
- Not sure about the PyCharm supports. I will test it and tell if it's possible to do what you want.
- There is no way currently to determine the number of splits automatically and to run them all. I think it's fine as it is for now.
- Regarding issue #454, it doesn't happen locally for me or on CI so I don't how to help with that. Perhaps @schroedk can help.

mdbenito (approved):
Thanks for the added docs. I think this is great. I think we can merge and release a patch version soon

mdbenito:
There seems to be some problem with "Run base tests"

mdbenito:
> * Not sure about the PyCharm supports. I will test it and tell if it's possible to do what you want.
> * There is no way currently to determine the number of splits automatically and to run them all. I think it's fine as it is for now.

I don't think this is necessary if the situation is as I described in [this comment](https://github.com/aai-institute/pyDVL/pull/429/files/03791ccedc4695fd47fabf668af167eaddffb863#r1374693395), namely:

* xdist is used locally
* split is used on CI only (or locally if manually started)

In this case, we get max core usage locally, and IDE built-in support to launch and track the parallel tests

AnesBenmerzoug:
@mdbenito `Run base tests` shouldn't exist anymore. It was replaced by `Run Tests - Python <version> - Group <group>` in this PR.

PR #431, opened 2023-09-15
Some improvements to OOB notebook

mdbenito:
### Description

This PR adds some text and supporting functions to the OOB notebook

### Changes

- 

### Checklist

- [ ] Updated Changelog


mdbenito:
@BastienZim I've worked a bit on your notebook, let me know if you have comments. I was a bit surprised by the very different results that one can obtain with different seeds, often obtaining a _degradation_ of performance with the removal of the worst 20% points. In the end I added complete randomization of the whole run including the splitting of the dataset to see what the true variance is. It is much greater, but things are predictable. I also added random seed handling to compute_oob and did a couple minor things here and there

src/pydvl/value/oob/oob.py
```diff
@@ -52,37 +58,53 @@ def compute_data_oob(
...
+    random_state = np.random.RandomState(rng.bit_generator)
 
     result: ValuationResult[np.int_, np.object_] = ValuationResult.empty(
         algorithm="data_oob", indices=u.data.indices, data_names=u.data.data_names
     )
 
     if is_classifier(u.model):
         bag = BaggingClassifier(
-            u.model, n_estimators=n_est, max_samples=max_samples, n_jobs=n_jobs
+            u.model,
```
AnesBenmerzoug, line 93: The utility object is not really used in this method. Perhaps we could define a `BaggingUtility` class? What do you think?

src/pydvl/value/oob/oob.py
```diff
@@ -109,32 +131,28 @@ def compute_data_oob(
     return result
 
 
-def point_wise_accuracy(preds: NDArray, y: NDArray) -> NDArray:
-    r"""Computes point wise accuracy
+def point_wise_accuracy(x1: NDArray[T], x2: NDArray[T]) -> NDArray[T]:
```
AnesBenmerzoug, line 134: I think we should use `y_true`, `y_pred` instead of `x1`, `x2` to follow the sklearn convention. It should be more familiar to users.

```suggestion
def point_wise_accuracy(y_true: NDArray[T], y_pred: NDArray[T]) -> NDArray[T]:
```

src/pydvl/value/oob/oob.py
```diff
@@ -109,32 +131,28 @@ def compute_data_oob(
...
+    r"""Point-wise 0-1 loss between two arrays
 
     Args:
-        preds: Model prediction on
-        y:  data labels corresponding to the model predictions
+        x1: Array of values (e.g. model predictions)
+        x2: Array of values (e.g. labels)
```
AnesBenmerzoug, line 139:
```suggestion
        y_true: Array of true values (e.g. labels)
        y_pred: Array of estimated values (e.g. model predictions)
```

src/pydvl/value/oob/oob.py
```diff
@@ -109,32 +131,28 @@ def compute_data_oob(
...
 
     Returns:
-        Array of point wise accuracy
+        Array with point-wise 0-1 losses between labels and model predictions
     """
-    return np.array(preds == y, dtype=np.int_)
+    return np.array(x1 == x2, dtype=x1.dtype)
```
AnesBenmerzoug, line 144:
```suggestion
    return np.array(y_pred == y_true, dtype=y_pred.dtype)
```

src/pydvl/value/oob/oob.py
```diff
@@ -109,32 +131,28 @@ def compute_data_oob(
...
 
 
-def neg_l2_distance(preds: NDArray[T], y: NDArray[T]) -> NDArray[T]:
-    r"""Computes negative l2 distance between label and model prediction
+def neg_l2_distance(x1: NDArray[T], x2: NDArray[T]) -> NDArray[T]:
```
AnesBenmerzoug, line 147:
```suggestion
def neg_l2_distance(y_true: NDArray[T], y_pred: NDArray[T]) -> NDArray[T]:
```

src/pydvl/value/oob/oob.py
```diff
@@ -109,32 +131,28 @@ def compute_data_oob(
...
+    r"""Point-wise negative $l_2$ distance between two arrays
 
     Args:
-        preds: Model prediction on
-        y:  data labels corresponding to the model predictions
+        x1: Array of values (e.g. model predictions)
+        x2: Array of values (e.g. labels)
```
AnesBenmerzoug, line 152:
```suggestion
        y_true: Array of true values (e.g. labels)
        y_pred: Array of estimated values (e.g. model predictions)
```

src/pydvl/value/oob/oob.py
```diff
@@ -109,32 +131,28 @@ def compute_data_oob(
...
+        Array with point-wise negative $l_2$ distances between labels and model
+        predictions
     """
-    return -np.square(
-        np.array(
-            preds - y,
-            dtype=np.float64,
-        )
-    )
+    return -np.square(np.array(x1 - x2), dtype=x1.dtype)
```
AnesBenmerzoug, line 158:
```suggestion
    return -np.square(np.array(y_pred - y_true), dtype=y_pred.dtype)
```

notebooks/data_oob.ipynb
```diff
@@ -43,167 +43,198 @@
...
-    "        \"education\",\n",
-    "        \"marital-status\",\n",
-    "        \"occupation\",\n",
-    "        \"relationship\",\n",
-    "        \"race\",\n",
-    "        \"sex\",\n",
-    "        \"native-country\",\n",
-    "    ]\n",
-    ")\n",
+    "The main idea of Data-OOB is to take a pre-trained classifier and compute a per-sample out-of-bag performance estimate via bagging.\n",
```
AnesBenmerzoug, line 221: I think there's a mistake here. It says that we're taking a pre-trained classifier but in the code and the method we trained one from scratch using bagging.

src/pydvl/value/oob/oob.py
mdbenito: The thing is that this is not tied to any model or prediction. It's just a point-wise metric. As such it doesn't make sense to use y_true and y_pred. But ok, I guess familiarity trumps rigour.

notebooks/data_oob.ipynb
mdbenito: Hmm... you are right, it's not "pre-trained", but "existing".

src/pydvl/value/oob/oob.py
```diff
@@ -98,7 +120,7 @@ def compute_data_oob(
     ):  # The bottleneck is the bag fitting not this part so TQDM is not very useful here
         oob_idx = np.setxor1d(u.data.indices, np.unique(samples))
         array_loss = loss(
-            preds=est.predict(u.data.x_train[oob_idx]), y=u.data.y_train[oob_idx]
+            x1=est.predict(u.data.x_train[oob_idx]), x2=u.data.y_train[oob_idx]
```
AnesBenmerzoug, line 123:
```suggestion
            y_true=u.data.y_train[oob_idx], y_pred=est.predict(u.data.x_train[oob_idx])
```

src/pydvl/value/oob/oob.py
mdbenito: I'm not sure, Utility was designed for margin-based methods and here it's being "abused" to conform to the interface. Evaluating this BaggingUtility on subsets could only mean computing the values themselves (i.e. the point-wise losses), which is a vector output and very different semantics from other utilities...

AnesBenmerzoug: You're right. We can leave it as is for now. We will eventually find a better interface.

AnesBenmerzoug (approved):
Looks good to me. It's only missing a changelog entry.

PR #437, opened 2023-09-22
Feature/filter converged

### LINKED ISSUE #303: Allow stopping criteria that operate on indices
As of 0.5.0 they are global: either a criterion is fulfilled for (a fraction of) the values or it is not. But often some values converge sooner than others, lower values taking typically longer. Once an index is tagged as converged, one can avoid computing any marginals for it.

mdbenito:
### Description

This PR closes #303 

### Changes

- Adds an additional parameter to `compute_generic_semivalues` that allows skipping samples for indices that have already converged
- Improves documentation and warns about that parameter (it's tricky)

Unrelated:
- Deprecates a custom attribute in `StoppingCriterion` in favour of `__str__`

### Checklist

- [x] Updated Changelog


kosmitive:
Can you elaborate why this parameter is necessary. Instead of doing that, we can also try to exclude marginal evaluations for converged indices, by removing the from the task queue.

mdbenito:
> Can you elaborate, why this parameter is necessary? Instead of doing that, we can also try to exclude marginal evaluations for converged indices, by removing them from the task queue.

This is exactly what is done. The parameter is necessary because just making it the default would change the behaviour for users. Also, very often values would converge only because of "unexpected" additional marginal computations.

This is something that happens often: we do `AbsoluteStandardError(0.02) | MaxUpdates(1000)`. Then some values might fulfill the first criterion really fast, despite being very bad estimates. Because the other indices take much longer to have low stderr and `done(result)` is a global check, the "converged" ones keep being updated and end up being good estimates. If we stopped updating them, the values would be off by a lot. The way to fix this is by making the stderr threshold much lower, e.g. `AbsoluteStandardError(1e-3) | MaxUpdates(1000)`. With `skip_converged=True` this more stringent check still takes less time than the first one.

src/pydvl/value/semivalues.py
```diff
@@ -262,16 +272,30 @@ def compute_generic_semivalues(
...
+                while len(pending) < n_submitted_jobs:
+                    samples = dict(islice(sampler_it, batch_size))
                     if len(samples) == 0:
                         raise StopIteration
 
-                    pending.add(
-                        executor.submit(
-                            _marginal, u=u, coefficient=correction, samples=samples
+                    # Filter out samples for indices that have already converged
+                    if skip_converged and len(done.converged) > 0:
```
kosmitive, line 281: Can be rewritten in functional form 

```python
filtered_samples = samples.items() 
if skip_converged and len(done.converged) > 0:
  filtered_samples = filter(lambda t: done.converged[t[0]], filtered_samples)
filtered_samples = tuple(filtered_samples)
```

even one-liner possible (but difficult to read).

src/pydvl/value/semivalues.py
```diff
@@ -198,6 +199,15 @@ def compute_generic_semivalues(
         coefficient: The semi-value coefficient
         done: Stopping criterion.
         batch_size: Number of marginal evaluations per single parallel job.
+        skip_converged: Whether to skip marginal evaluations for indices that
```
kosmitive, line 202: How about printing a warning as well?

src/pydvl/value/stopping.py
```diff
@@ -239,9 +287,8 @@ def converged(self) -> NDArray[np.bool_]:
                 return super().converged
             return converged()
 
-        @property
-        def name(self):
-            return self._name
+        def __str__(self):
+            return f"{self._name}"
```
kosmitive, line 291: Better without f-string replacement operator:

```python 
def __str__(self):
    return self._name
```

tests/value/test_semivalues.py
```diff
@@ -48,12 +48,13 @@ def test_shapley(
...
 ):
     u, exact_values = analytic_shapley
-    criterion = AbsoluteStandardError(0.02, 1.0) | MaxUpdates(2 ** (num_samples * 2))
+    criterion = AbsoluteStandardError(1e-4) | MaxUpdates(1000)
     values = compute_generic_semivalues(
         sampler(u.data.indices),
         u,
         coefficient,
         criterion,
+        skip_converged=True,
```
kosmitive, line 57: Case `skip_converge=False` missing. Maybe better to include a specific test case for skip converged.

kosmitive (changes requested):

src/pydvl/value/semivalues.py
mdbenito: 🤔 About what? There is no change in behaviour from previous versions. Or do you mean we should transition to a default of `true` and warn that this will be so?

tests/value/test_semivalues.py
mdbenito: Hmmm... maybe. I didn't want to add more tests. In a sense it contains skip=True as a special case where no indices converge before all do (yes, that's a stretch)

mdbenito (commented):
Thanks @kosmitive for the comments. Implementing a couple

src/pydvl/value/semivalues.py
kosmitive: I mean a warning if ones uses a different Stopping Criterion than advised.

mdbenito: No, I don't think so. This parameter doesn't change what's recommended or isn't, it only highlights an issue with global stopping criteria. There should be good documentation about what the different criteria do, but no warnings annoying people who possibly know what they are doing.

mdbenito:
> This is something that happens often: we do `AbsoluteStandardError(0.02) | MaxUpdates(1000)`. Then some values might fulfill the first criterion really fast, despite being very bad estimates. Because the other indices take much longer to have low stderr and `done(result)` is a global check, the "converged" ones keep being updated and end up being good estimates. If we stopped updating them, the values would be off by a lot. The way to fix this is by making the stderr threshold much lower, e.g. `AbsoluteStandardError(1e-3) | MaxUpdates(1000)`. With `skip_converged=True` this more stringent check still takes less time than the first one.

Added something along these lines to the docs for `stopping.py`

PR #439, opened 2023-09-22
Improve antithetic sampling

mdbenito:
### Description

This PR closes fixes a bug in antithetic sampling and adds support for permutation antithetic sampling

### Changes

- When returning a complement set, do not add the index to the set since this results in no marginal being computed
- Added `AntitheticPermutationSampler`

### Checklist

- [x] Updated Changelog

### DIFF
--- a/src/pydvl/value/sampler.py
+++ b/src/pydvl/value/sampler.py
@@ -315,18 +315,19 @@ class AntitheticSampler(StochasticSamplerMixin, PowersetSampler[IndexT]):
     """An iterator to perform uniform random sampling of subsets, and their
     complements.
 
-    Works as :class:`~pydvl.value.sampler.UniformSampler`, but for every tuple
-    $(i,S)$, it subsequently returns $(i,S^c)$, where $S^c$ is the complement of
-    the set $S$, including the index $i$ itself.
+    Works as [UniformSampler][pydvl.value.sampler.UniformSampler], but for every
+    tuple $(i,S)$, it subsequently returns $(i,S^c)$, where $S^c$ is the
+    complement of the set $S$ in the set of indices, excluding $i$.
     """
 
     def __iter__(self) -> Iterator[SampleT]:
         while True:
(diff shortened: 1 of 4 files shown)


src/pydvl/value/sampler.py
```diff
@@ -372,6 +382,27 @@ def weight(cls, n: int, subset_len: int) -> float:
...
+    href="#mitchell_sampling_2022">1</a></sup>
+    """
+
+    def __iter__(self) -> Iterator[SampleT]:
+        while True:
+            permutation = self._rng.permutation(self._indices)
+            for perm in permutation, permutation[::-1]:
+                for i, idx in enumerate(perm):
+                    yield idx, perm[:i]
+                    self._n_samples += 1
```
AnesBenmerzoug, line 400: Shouldn't we yield the reversed permutation right after the first one?
If I'm not mistaken, this implementation goes through the entire permutation first before yielding the reversed ones.  

Perhaps this would be better?
```suggestion
            permutation = self._rng.permutation(self._indices)
            for i, idx in enumerate(permutation):
                 yield idx, permutation[:i]
                 self._n_samples += 1
                 yield idx, permutation[i+1:][::-1]
                 self._n_samples += 1
```

For the array of indices `indices = np.array([1, 2, 3])` this should generate (ignoring the permutation):

```python
1 []
1 [3, 2]
2 [1]
2 [3]
3 [1, 2]
3 []
```

mdbenito: Yes, it does one full permutation before the other one. This is how it's described in the paper which first described antithetic permutation sampling. It's the same as with sets: first one set, then the complement. In this case a permutation, then its reverse. We could also start from both ends. Each of these has different properties that I haven't investigated at all, so I was content with doing what the paper describes, but all this probably warrants documenting and warning.

mdbenito: See #442

PR #440, opened 2023-09-22
Use pytest-xdist

mdbenito:
### Description

This PR makes our lives nicer by using pytest-xdist. We automagically set the number of pytest workers to avoid oversubscription by looking at `num_workers` (the number of parallel processes started within a test using a parallel backend) 

### Changes

- Required pytest-xdist
- Made `num_workers` not a fixture
- Used `-n auto` in calls to pytest inside `tox.ini`

### Checklist

- [ ] ~Updated Documentation (if necessary)~
- [x] Updated Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Using pytest-xdist for faster local tests
+  [PR #440](https://github.com/aai-institute/pyDVL/pull/440)
 - Implementation of Data-OOB by @BastienZim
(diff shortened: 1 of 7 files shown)


AnesBenmerzoug:
This doesn't seem to improve the test runtime on CI. Is that expected?

AnesBenmerzoug (approved):

mdbenito:
> This doesn't seem to improve the test runtime on CI. Is that expected?

For the record: yes, this only improves tests on machines with > 4 cores.

PR #456, opened 2023-11-19
Feature/377 lazy tensor

### LINKED ISSUE #26: Parallelize IF impl. (by samples)
We want:
* To split the computation of the influence of training points across processors and machines
* To reuse computations (e.g. in #24 we might want to reuse the samples across processes, if this makes sense)

### LINKED ISSUE #277: Parallelise influence calculation
There are a few loops in the influence code that may use the speed up from parallelisation:
- in TorchDifferentiable.split_grad and TorchDifferentiable.mvp
- in _calculate_influences_pert, within general.py
- in conjugate_gradient, within inversion_methods
Since they all do backpropagation under the hood, memory and not compute may be the biggest bottleneck in practice. Regardless, including the possibility to parallelise should be quite straightforward. 

Related to #26

### LINKED ISSUE #377: Persist IF tensors to storage for large models
Problem: As highlighted in [a discussion](https://github.com/appliedAI-Initiative/pyDVL/pull/375#pullrequestreview-1527972934) in MR #375, a lot of the tensors we create for influence calculation end up being very large. For example, the influence factors are stored in a matrix of size (N_test_points, N_model_parameters). Often, these cannot fit into RAM (be it ordinary or on GPU). 

A simple fix would use generators. Instead of creating all the influence factors, we could calculate batch-by-batch within the compute_influences_up and compute_influences_pert methods. Same can be said of grads within these methods. However, this would imply re-calculating influence factors or grads several times, with massive computational overhead.

A better fix would cache the tensors to disk. We could create a CachedTensor class that saves the tensors into files and loads them whenever needed, thus freeing the VRAM. Even the output of the compute_influences method could be a CachedTensor: one would only need to remember to flush it when it is no longer needed.

### LINKED ISSUE #430: Update, fix and improve Influence Functions notebooks
- [x] Apply tags to hide import cells.
- [x] Fix rendering of certain math equations in influence imagenet notebook.
- [ ]  ~~Consider using pytorch lightning instead of custom TrainingManager class. See #351~~
- [x] Remove extra empty cell in influence synthetic notebook.
- [x] Fix usage of `support` module

### LINKED ISSUE #457: Fix import issue in README.md example
There is an import error for TorchTwiceDifferentiable in the current README example.
In addition, model not in eval mode warning is shown.

schroedk:
### Description

This PR closes #377 closes #430 closes #26 closes #457 closes #277

### Added
- New influence function interface `InfluenceFunctionModel`
- Data parallel computation with `DaskInfluenceCalculator`
  [PR #26](https://github.com/aai-institute/pyDVL/issues/26)
- Sequential batch-wise computation and write to disk with `SequentialInfluenceCalculator` 
  [PR #377](https://github.com/aai-institute/pyDVL/issues/377)
- Adapt notebooks to new influence abstractions
  [PR #430](https://github.com/aai-institute/pyDVL/issues/430)

### Changed 
- **Breaking Changes**
  - Removed `compute_influences` and all related code.
    Replaced by new `InfluenceFunctionModel` interface. Removed modules:
    - influence.general
    - influence.inversion
    - influence.twice_differentiable
    - influence.torch.torch_differentiable

### Fixed
- Import bug in README [PR #457](https://github.com/aai-institute/pyDVL/issues/457)

### Checklist

- [x] Updated Changelog


src/pydvl/influence/influence_calculator.py
```diff
@@ -0,0 +1,554 @@
...
+    InfluenceFunctionModel,
+    InfluenceType,
+    TensorType,
+    UnSupportedInfluenceTypeException,
+)
+
+__all__ = ["DaskInfluenceCalculator"]
+
+
+class DimensionChunksException(ValueError):
```
mdbenito, line 20: Isn't this more natural?
```suggestion
class ChunksDimensionsException(ValueError):
```

src/pydvl/influence/influence_calculator.py
```diff
@@ -0,0 +1,554 @@
...
+        except ValueError:
+            return None
+
+
+class BlockAggregator(Generic[TensorType], ABC):
+    @abstractmethod
+    def aggregate_nested(
+        self, tensors: Generator[Generator[TensorType, None, None], None, None]
+    ):
+        """Overwrite this method to aggregate provided blocks into a single tensor"""
```
mdbenito: The docstring doesn't help me understand what the input is like (nested?), and what I should do with it. But I think this function is just a "flatten", right? Then it should be called flatten

src/pydvl/influence/influence_calculator.py
```diff
@@ -0,0 +1,554 @@
...
+
+    @abstractmethod
+    def aggregate(self, tensors: Generator[TensorType, None, None]):
```
mdbenito: Isn't this a concatenation? I would name it `concat` then

src/pydvl/influence/influence_calculator.py
```diff
@@ -0,0 +1,554 @@
...
+
+    @staticmethod
+    def _get_client() -> Optional[distributed.Client]:
+        try:
+            return distributed.get_client()
+        except ValueError:
+            return None
+
+
+class BlockAggregator(Generic[TensorType], ABC):
```
mdbenito: What's a "block"? Why not `TensorAggregator`?

src/pydvl/influence/torch/functional.py
```diff
@@ -1,27 +1,62 @@
-from typing import Callable, Collection, Dict, Generator, Iterable, Mapping, Tuple, cast
+"""
+This module provides methods for efficiently compute tensors related to first and second order derivatives
```
mdbenito, line 2:
```suggestion
This module provides methods for efficiently computing tensors related to first and second order derivatives
```

src/pydvl/influence/torch/functional.py
```diff
@@ -62,70 +97,78 @@ def hvp(
...
 
-    Yields:
-        Partial functions `H_{batch}(vec)=hvp(model, loss, inputs, targets, vec)` that when called,
-            will compute the Hessian-vector product H(vec) for the given model and loss in a batch-wise manner, where
-            (inputs, targets) coming from one batch.
-    """
+    Returns:
+        Callable[[torch.Tensor, torch.Tensor, torch.Tensor], torch.Tensor]: A function that takes three torch.Tensor
+        objects - input data (x), target data (y), and a vector (vec) - and returns the Hessian-vector product as a
+        torch.Tensor.
```
mdbenito, line 125: No need to repeat types in the docstrings.
```suggestion
        A function that takes three `torch.Tensor` objects - input data (`x`), target data (`y`), and a vector
        (`vec`), and returns the Hessian-vector product of the loss evaluated on `x`, `y` times `vec`.
```

src/pydvl/influence/torch/functional.py
```diff
@@ -1,27 +1,62 @@
...
-    params: TorchTensorContainerType,
-    vec: TorchTensorContainerType,
+    func: Callable[[Dict[str, torch.Tensor]], torch.Tensor],
+    params: Dict[str, torch.Tensor],
+    vec: Dict[str, torch.Tensor],
     reverse_only: bool = True,
-) -> TorchTensorContainerType:
+) -> Dict[str, torch.Tensor]:
     r"""
     Computes the Hessian-vector product (HVP) for a given function at given parameters, i.e.
```
mdbenito, line 61:
```suggestion
    Computes the Hessian-vector product (HVP) for a given function at the given parameters, i.e.
```

src/pydvl/influence/torch/functional.py
```diff
@@ -201,55 +248,540 @@ def get_hvp_function(
...
-        use_hessian_avg: If True, the returned function uses batch-wise Hessian computation via
+        precompute_grad: If True, the full data gradient is precomputed and kept in memory, which can speed up
+            the hessian vector product computation. Set this to False, if you can't afford to keep an additional
+            parameter-sized vector in memory.
+        use_average: If True, the returned function uses batch-wise computation via
             [batch_loss_function][pydvl.influence.torch.functional.batch_loss_function] and averages the results.
             If False, the function uses backpropagation on the full
             [empirical_loss_function][pydvl.influence.torch.functional.empirical_loss_function],
             which is more accurate than averaging the batch hessians, but probably has a way higher memory usage.
         reverse_only: Whether to use only reverse-mode autodiff (True, default) or
```
mdbenito, line 259: Types and default values are redundant in docstrings (there are many instances, just pointing this one out)

```suggestion
        reverse_only: Whether to use only reverse-mode autodiff or
```

mdbenito (commented):
I added some superficial comments about docstrings and names here and there. I don't understand all of this enough yet to say something substantial, sorry

src/pydvl/influence/influence_calculator.py
schroedk: Renamed it to InvalidDimensionChunksError

src/pydvl/influence/influence_calculator.py
```diff
@@ -0,0 +1,554 @@
...
+
+    def __init__(
+        self,
+        influence_function_model: InfluenceFunctionModel,
+        numpy_converter: NumpyConverter,
+    ):
+        self._num_parameters = influence_function_model.num_parameters
+        self.influence_function_model = influence_function_model
+        self.numpy_converter = numpy_converter
+        client = self._get_client()
```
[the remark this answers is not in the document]
schroedk: The only way to ensure, that the user uses the distributed client for computing, is to take the client as an input parameter to the init. I documented it in the class as well as in the documentation page. I will create an issue for this, so we can discuss the appropriate way to handle it.

src/pydvl/influence/influence_calculator.py
```diff
@@ -0,0 +1,554 @@
...
+        influence_function_model: instance of type
+            [InfluenceFunctionModel][pydvl.influence.base_influence_model.InfluenceFunctionModel], defines the
+            batch-wise computation model
+        numpy_converter: instance of type [NumpyConverter][pydvl.influence.influence_calculator.NumpyConverter], used to
+            convert between numpy arrays and TensorType objects needed to use the underlying model
+    """
+
+    def __init__(
+        self,
+        influence_function_model: InfluenceFunctionModel,
```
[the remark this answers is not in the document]
schroedk: Actually, I disagree that less is better. To quote a colleague of ours: Clarity over brevity. What would be the advantage to name it if_model?

docs/influence/index.md
```diff
@@ -477,25 +350,95 @@ where \(D\) is a diagonal matrix with the top (in absolute value) eigenvalues of
...
+In case, you need to stick to a sequential processing, set `n_workers` accordingly
+```python
+client = Client(n_workers=1, threads_per_worker=1) 
+```
+
+!!! Warning
+    Make sure, you set `threads_per_worker=1`, due to some torch.autograd functions not being thread-safe
+
+
+
```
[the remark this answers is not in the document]
schroedk: Made the warning more prominent and also add it to the class docstring.

src/pydvl/influence/influence_calculator.py
schroedk: Depends on the implementation. I made the docstring more verbose, please check again.

src/pydvl/influence/influence_calculator.py
```diff
@@ -0,0 +1,554 @@
...
+    def from_numpy(self, x: NDArray) -> TensorType:
+        """Overwrite this method for converting a numpy array into a TensorType object"""
+
+
+class DaskInfluenceCalculator:
+    """
+    Compute influences over dask.Array collections. Depends on a batch computation model
+    of type [InfluenceFunctionModel][pydvl.influence.base_influence_model.InfluenceFunctionModel].
+    In addition, provide transformations from and to numpy,
+    corresponding to the tensor types of the batch computation model.
```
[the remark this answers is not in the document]
schroedk: Improved docstring, please check again

src/pydvl/influence/influence_calculator.py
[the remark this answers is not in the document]
Xuzzo: I also think longer names are better as they are more descriptive. When I read just "model" I think it is only the NN

Xuzzo:
Looks good to me. @mdbenito Could we go ahead and merge?

src/pydvl/influence/array.py
```diff
@@ -0,0 +1,390 @@
...
+class LazyArrayProvider(ABC):
+    """
+    Abstract base class representing a provider of lazily evaluated, array-like data. This class
+    is tailored for situations where data is not immediately computed or loaded, but is instead
+    generated or processed on-demand. The 'array-like' aspect refers to the data structure or
+    format rather than the ability to perform typical array operations like indexing.
+    """
+
+    @abstractmethod
+    def compute(self, tensor_aggregator: Optional[TensorAggregator] = None):
```
mdbenito, line 127: I think this is another example of an IMO unnecessary type prefix. It's an aggregator of type `TensorAggregator`. As a user, I don't need to see a name `tensor_aggregator`, that's just only going to take up space.

src/pydvl/influence/array.py
```diff
@@ -0,0 +1,390 @@
...
+            The result of the computation or data retrieval, processed as specified by the
+            tensor_aggregator, if provided. The exact return type and format depend on the
+            implementation and the aggregator used.
+        """
+
+    @abstractmethod
+    def to_zarr(
+        self,
+        path_or_url: str,
+        numpy_converter: NumpyConverter,
```
mdbenito, line 147: And again: the only type supported is `NumpyConverter` => don't need it in the name.

src/pydvl/influence/array.py
```diff
@@ -0,0 +1,390 @@
...
+    """
+    An aggregator that collects tensors into list structures.
+    """
+
+    def aggregate_from_nested_generators(
+        self,
+        nested_generators_of_tensors: Generator[
+            Generator[TensorType, None, None], None, None
+        ],
+    ):
```
mdbenito, line 88: Missing return type.

src/pydvl/influence/array.py
```diff
@@ -0,0 +1,390 @@
...
+    methods for aggregating tensors.
+    """
+
+    @abstractmethod
+    def aggregate_from_nested_generators(
+        self,
+        nested_generators_of_tensors: Generator[
+            Generator[TensorType, None, None], None, None
+        ],
+    ):
```
mdbenito, line 54: Missing return type

src/pydvl/influence/array.py
```diff
@@ -0,0 +1,390 @@
...
+            Depending on the value of return_stored, this method may return a reference to the stored
+            Zarr data object or None.
+        """
+
+
+class OneDimChunkedLazyArrayProvider(LazyArrayProvider):
+    """
+    A class representing chunked, and lazily evaluated array, where the chunking is restricted to the first dimension
+
+    This class is designed to handle large arrays that are not feasible to load into memory all at once.
```
mdbenito, line 173:
```suggestion
    This class is designed to handle large arrays that don't fit in memory.
```

src/pydvl/influence/array.py
```diff
@@ -0,0 +1,390 @@
...
+
+        Returns:
+            Depending on the value of return_stored, this method may return a reference to the stored
+            Zarr data object or None.
+        """
+
+
+class OneDimChunkedLazyArrayProvider(LazyArrayProvider):
+    """
+    A class representing chunked, and lazily evaluated array, where the chunking is restricted to the first dimension
```
mdbenito, line 171:
```suggestion
    A class representing a chunked, and lazily evaluated array, where the chunking is restricted to the first dimension
```

Also, please remember the column limit.

src/pydvl/influence/array.py
```diff
@@ -0,0 +1,390 @@
...
+        Args:
+            tensor_generator: A generator that yields TensorType objects.
+
+        Returns:
+            A list containing all the tensors provided by the tensor_generator.
+        """
+        return [t for t in tensor_generator]
+
+
+class LazyArrayProvider(ABC):
```
mdbenito, line 118: I don't think this is a "Provider" in the standard sense. There is no global state, nor centralised management of resources nor anything like this. If anything this is a `LazyArrayEvaluator` (please don't use that name 😅). If we don't care about the overloading of the name "Provider" from the design pattern, then maybe `ArrayProvider`, since it is arrays that are provided. But in fact this is a "Realizer" or "Evaluator".

src/pydvl/influence/array.py
```diff
@@ -0,0 +1,390 @@
...
+            tensor_aggregator = ListAggregator()
+        return tensor_aggregator.aggregate_from_generator(self.generator_factory())
+
+    def to_zarr(
+        self,
+        path_or_url: str,
+        numpy_converter: NumpyConverter,
+        return_stored: bool = False,
+        overwrite: bool = False,
+    ):
```
mdbenito, line 210: Missing return type

src/pydvl/influence/array.py
```diff
@@ -0,0 +1,390 @@
...
+
+        Args:
+            path_or_url: The file path or URL where the Zarr array will be stored.
+            numpy_converter: A converter for transforming blocks into NumPy arrays compatible with Zarr.
+            return_stored: If True, the method returns the stored Zarr array; otherwise, it returns None.
+            overwrite: If True, overwrites existing data at the given path_or_url. If False, an error
+                is raised in case of existing data.
+
+        Returns:
+            [zarr.core.Array][zarr.core.Array] or None: The Zarr array if return_stored is True; otherwise, None.
```
mdbenito, line 224: We should specify types in signatures, not docstrings

src/pydvl/influence/array.py
```diff
@@ -0,0 +1,390 @@
...
+        """
+        Converts the array into Zarr format, a storage format optimized for large arrays, and stores it at
+        the specified path or URL. This method is suitable for scenarios where the data needs to be saved
+        for later use or for large datasets requiring efficient storage.
+
+        Args:
+            path_or_url: The file path or URL where the Zarr array will be stored.
+            numpy_converter: A converter for transforming blocks into NumPy arrays compatible with Zarr.
+            return_stored: If True, the method returns the stored Zarr array; otherwise, it returns None.
+            overwrite: If True, overwrites existing data at the given path_or_url. If False, an error
```
mdbenito, line 220: Formatting of variable names and values (here and elsewhere)
```suggestion
            overwrite: If `True`, overwrites existing data at the given `path_or_url`. If `False`, an error
```

src/pydvl/influence/base_influence_model.py
```diff
@@ -0,0 +1,173 @@
...
+
+class InfluenceType(str, Enum):
+    r"""
+    Enum representation for the types of influence.
+
+    Attributes:
+        Up: Up-weighting a training point, see section 2.1 of
+            (Koh and Liang, 2017)<sup><a href="#koh_liang_2017">1</a></sup>
+        Perturbation: Perturb a training point, see section 2.2 of
+            (Koh and Liang, 2017)<sup><a href="#koh_liang_2017">1</a></sup>
```
mdbenito, line 14: Don't we have this in our documentation? It's always better to link to it. Going to a paper for details is very cumbersome

tests/influence/conftest.py
```diff
@@ -195,3 +196,57 @@ def add_noise_to_linear_model(
...
+def analytical_linear_influences(
+    linear_model: Tuple[NDArray[np.float_], NDArray[np.float_]],
+    x: NDArray[np.float_],
+    y: NDArray[np.float_],
+    x_test: NDArray[np.float_],
+    y_test: NDArray[np.float_],
+    influence_type: InfluenceType = InfluenceType.Up,
+    hessian_regularization: float = 0,
+):
+    """Calculates analytically the influence of each training sample on the
```
mdbenito, line 210: Old sphinx docstring format (I know it's not rendered, but still)

src/pydvl/influence/base_influence_model.py
```diff
@@ -0,0 +1,173 @@
...
+
+    @property
+    @abstractmethod
+    def n_parameters(self):
+        """Number of trainable parameters of the underlying model"""
+
+    @property
+    @abstractmethod
+    def is_fitted(self):
+        """Overwrite this, to expose the fitting status of the instance."""
```
mdbenito, line 57: I think you mean "override" (also: German comma ;) ). However, if an overriden method doesn't define a doscstring, the ide will show this one and that would be confusing. What about:
```suggestion
        """Whether or not the influence model has been fitted."""
```

src/pydvl/influence/base_influence_model.py
```diff
@@ -0,0 +1,173 @@
...
+
+    @abstractmethod
+    def fit(self, data_loader: DataLoaderType):
+        """
+        Overwrite this method to fit the influence function model to training data, e.g. pre-compute hessian matrix
```
mdbenito, line 62: "Override" (and elsewhere too). Missing preposition and article
```suggestion
        Override this method to fit the influence function model to training data, e.g. to pre-compute the hessian matrix
```
Also, please format to 88 cols like the rest of the code

src/pydvl/influence/base_influence_model.py
```diff
@@ -0,0 +1,173 @@
...
+
+        """
+
+    def influences(
+        self,
+        x_test: TensorType,
+        y_test: TensorType,
+        x: Optional[TensorType] = None,
+        y: Optional[TensorType] = None,
+        influence_type: InfluenceType = InfluenceType.Up,
```
mdbenito, line 101: What other type than `influence` could it be? This should be `type`. Because that's a reserved keyword, what about `mode` or `kind`? note that `InfluenceType` is also strange because it's not a type in the software sense, so that `InfluenceMode` makes more sense IMO

src/pydvl/influence/influence_calculator.py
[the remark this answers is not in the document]
mdbenito: But the question is whether the client object here has the number of threads set to 1. Not whether the user is using some type of client or another.

Also: what issue?

src/pydvl/influence/influence_calculator.py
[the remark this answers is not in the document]
mdbenito: The problem is consistency. We can't prefix some variable names with types and not others. Since I agree with the confusion with the NN model, I propose to use `if_model` as a compromise, and to use the prefix `if_` consistently. 
Longer variable and class names are useful, I agree. But as I explain in another comment, they can be a mental drain and just get in the way. And it's not an either/or situation here (I'm not suggesting to call it `m`)

mdbenito (commented):
More comments about naming, docs, types. Sorry. Happy to discuss in a call. I'm also a bit confused about not seeing a way to process the chunks in sequence without reading them into memory or a file all at once.

src/pydvl/influence/array.py
```diff
@@ -0,0 +1,390 @@
...
+
+Classes:
+    NumpyConverter: Abstract base class for converting between tensor types and NumPy arrays.
+    BlockAggregator: Abstract base class for aggregating blocks of data.
+    ListAggregator: Concrete implementation of BlockAggregator using lists.
+    LazyArray: Abstract base class representing a lazily evaluated array.
+    OneDimChunkedLazyArray: Represents a chunked and lazily evaluated array, where chunking is restricted to the
+        first dimension.
+    TwoDimChunkedLazyArray: Represents a chunked and lazily evaluated array, where chunking is restricted to the
+        two first dimension.
```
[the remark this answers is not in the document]
schroedk: Agree, will remove it

src/pydvl/influence/array.py
```diff
@@ -0,0 +1,390 @@
...
+
+
+class TensorAggregator(Generic[TensorType], ABC):
+    """
+    Abstract base class for aggregating tensors. Implementations of this class should define
+    methods for aggregating tensors.
+    """
+
+    @abstractmethod
+    def aggregate_from_nested_generators(
```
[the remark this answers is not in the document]
schroedk: An aggregator is something a user can provide. For example, if the user is only interested in the mean she/he could do
`class TorchMeanAggregator(TensorAggregator):

    def aggregate_from_generator(self, tensor_generator: Generator[torch.Tensor, None, None]):
        return torch.mean(torch.cat([torch.mean(t) for t in tensor_generator]))

...
`
So there is only ever one complete chunk of data in memory. If we call the method flatten, this would be very confusing in my opinion.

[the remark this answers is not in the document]
mdbenito: That makes sense. It's the listaggregator below that's flattening. Understood.

src/pydvl/influence/base_influence_model.py
schroedk: Agree, InfluenceKind is better, will change it to
```
kind: InfluenceKind
```

src/pydvl/influence/base_influence_model.py
schroedk: Yes, override is correct....
This means black is not working for the docstrings?

src/pydvl/influence/array.py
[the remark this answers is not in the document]
mdbenito: Now that I understand this better (hopefully), I see it's just an interface for map-reduce of sorts. There are two steps, correct?

1. Process a structure of nested generators into one generator. (flatten_generators)
2. Process each yielded chunk in some way. (reduce)

Is there a need for two separate functions when it's apparently the same (recursive) process? And could we provide a helper method to produce a generator that "flattens" nested generators? Then users only need implement one method.

So in the end this is just one function `reduce_tensors(reduction_fun, tensors)`, which could actually be `functools.reduce` in some cases, if we provide the generator flattening (although that might be memory inefficient with all the `cat()`s...). I'm just spitballing here.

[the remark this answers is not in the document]
schroedk: Actually, we don't want to flatten the nested generators (which you could easily do with itertools.chain), because
it resembles the block matrix structure (every inner generator yields the blocks for one row group) and you would loose this representation, if you just flatten the generator. But it might makes sense to split the two functions into different classes, because I never need both methods at the same time. Additionally, the purpose of the class is clearer. But even in this case, it might not be a reduction (list, torch.cat,.....).

README.md
```diff
@@ -155,30 +152,38 @@ For influence computation, follow these steps:
...
+   influences = influence_function_model.influences(test_x, test_y, train_x, train_y)
    ```
    The result is a tensor of shape `(training samples x test samples)`
    that contains at index `(i, j`) the influence of training sample `i` on
    test sample `j`.
 
+7. For larger data, wrap the model into a
+   calculator and call methods on the calculator.
+   ```python
+   sequential_influence_calculator = SequentialInfluenceCalculator(influence_function_model)
```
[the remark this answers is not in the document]
schroedk: Check the notebooks and documentation for changing names

src/pydvl/influence/array.py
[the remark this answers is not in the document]
mdbenito: Understood, thx

src/pydvl/influence/array.py
```diff
@@ -0,0 +1,390 @@
...
+
+        Implement this method to specify how tensors, nested in multiple layers of generators,
+        should be combined. Useful for complex data structures where tensors
+        are not directly accessible in a flat list.
+
+        Args:
+            nested_generators_of_tensors: A generator of generators, each yielding TensorType objects."""
+
+    @abstractmethod
+    def aggregate_from_generator(
```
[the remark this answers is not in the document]
mdbenito: I was wrong!

src/pydvl/influence/array.py
```diff
@@ -0,0 +1,390 @@
...
+        tensors are yielded one by one.
+        """
+
+
+class ListAggregator(TensorAggregator):
+    """
+    An aggregator that collects tensors into list structures.
+    """
+
+    def aggregate_from_nested_generators(
```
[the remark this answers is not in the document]
mdbenito: Wrong.

src/pydvl/influence/influence_calculator.py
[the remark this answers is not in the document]
mdbenito: As discussed, we leave it as it is.

src/pydvl/influence/influence_calculator.py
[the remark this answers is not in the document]
mdbenito: As discussed, we make the client an argument and allow for explicit disabling of the check.

PR #458, opened 2023-11-23
Create abstraction for caching

### LINKED ISSUE #124: Offer alternative to memcache outside docker
Our data shapley methods heavily rely on memcache. On windows, the best (only?) way to use memcache is through Docker, which is a problem cause Docker is not free of charge. 
In the spirit of making our repository plug and play and as independent as possible, we should offer an alternative to memcache within our utility wrapper.

### LINKED ISSUE #189: Create abstraction for `Cache`
We should create a proper abstraction of the caching to make the code base cleaner and to allow alternative implementations.

We currently only support  `Memcached`.

This is related to issue #124

AnesBenmerzoug:
### Description

This PR closes #189 closes #124 

I tried at first to rely on joblib's Memory class by just implementing a new backend for Memcached but that proved to be too cumbersome because it relies on I/O operations (opening files, writing to files, etc.) so I just took inspiration from their versions while keeping some details from our previous implementation (e.g. CacheStats, repeated evaluations).

I created a separate issue (#459) for the creation of a notebook showcasing the use of caching and its benefits.

### Changes

- Replace caching module with caching package.
- Create `CacheBackend` base class for caching backend implementations.
- Implement `InMemoryCacheBackend`, `DiskCacheBackend`, `MemcachedCacheBackend`.
- Create `CachedFunc` class to wrap cached functions and methods.
- Change default caching `time_threshold` from 0.3 to 0.0
- Create new `memcached` extra and thus make `pymemcache` an optional dependency.
- Rename config class `MemcachedConfig` to `CachedFuncConfig`, remove memcached client config from it.
- Move caching configuration to separate config module inside the caching package.
- Adapt Utility to caching changes.
- Update existing tests and add new ones.
- Update and improve installation documentation with a clear definition and description of all the extra dependencies.
- Move all documentation related to caching, using memcached and parallelization to the first-steps page.
- Remove caching section from readme.
- Add link to extras section in documentation to readme.

### Checklist

- [x] Updated Changelog


mdbenito:
A high-level question: what is the use-case for the in-memory cache if it doesn't use [inter-process shared memory](https://docs.python.org/3.8/library/multiprocessing.shared_memory.html) or, much easier, a [multiprocessing manager](https://docs.python.org/3/library/multiprocessing.html#managers)?

It kicks in when submitting batches of samples to a worker, correct? But what about single samples?

AnesBenmerzoug:
@mdbenito Yes, it would be used in a single step of computations in a worker e.g. a single permuation.
It is only useful really for permutation based shapley computation.
I see the `InMemoryCacheBackend` as very similary to `@lru_cache`. So that if we ever deem the caching system to not be that useful we can just decide to use the latter directly.

src/pydvl/utils/caching/config.py
```diff
@@ -0,0 +1,40 @@
...
+            [memcached()][pydvl.utils.caching.memcached] will stop evaluating the function once the
+            standard deviation of the mean is smaller than `rtol_stderr * mean`.
+        min_repetitions: minimum number of times that a function evaluation
+            on the same arguments is repeated before returning cached values. Useful
+            for stochastic functions only. If the model training is very noisy, set
+            this number to higher values to reduce variance.
+    """
+
+    ignore_args: Collection[str] = field(default_factory=list)
+    time_threshold: float = 0
```
mdbenito, line 37: But time thresholds should depend on the latency of the caching backend, shouldn't they? We definitely don't want to cache every function call to memcached and definitely not to disk, so that a 0 here is not good for the users.

The default 0.3 was a hacky way of avoiding that, agreed. Instead what we need is an estimate of the latency in the backend, and the time of computation for the wrapped function . If in the first call the latter takes longer than the latency by a factor of, say 2, then we store to cache and upon the next call, check these numbers again and retrieve from cache. Backends can define a default latency or, ideally, they can precompute it dynamically upon construction by averaging a few hundred dummy read/writes.

I think we should do this since it avoids a major potential performance issue and just takes a couple of calls to `time()`

src/pydvl/utils/utility.py
```diff
@@ -134,8 +135,8 @@ def __init__(
         score_range: Tuple[float, float] = (-np.inf, np.inf),
         catch_errors: bool = True,
         show_warnings: bool = False,
-        enable_cache: bool = False,
-        cache_options: Optional[MemcachedConfig] = None,
+        cache_backend: Optional[CacheBackend] = None,
+        cached_func_options: CachedFuncConfig = CachedFuncConfig(),
```
mdbenito, line 139: Default value is mutable. I would either freeze the dataclass in the declaration or use None and a check below

src/pydvl/utils/caching/disk.py
```diff
@@ -0,0 +1,116 @@
...
+    ) -> None:
+        """Initialize the disk cache backend.
+
+        Args:
+            cache_dir: Base directory for cache storage.
+                By default, this is set to `~/.pydvl_cache/disk`
+        """
+        super().__init__()
+        self.cache_dir = Path(cache_dir)
+        self.cache_dir.mkdir(exist_ok=True, parents=True)
```
mdbenito, line 78: Wouln't it be better (and more portable) to use [tempfile](https://docs.python.org/3/library/tempfile.html)?

src/pydvl/utils/caching/memcached.py
```diff
@@ -0,0 +1,205 @@
...
+        ```
+    """
+
+    def __init__(self, config: MemcachedClientConfig = MemcachedClientConfig()) -> None:
+        """Initialize memcached cache backend.
+
+        Args:
+            config: Memcached client configuration.
+        """
+        if not PYMEMCACHE_INSTALLED:
```
mdbenito, line 106: I know it's trivial, but maybe we should use pytest-mock to test this? (The pattern probably repeats elsewhere)
```python

from pydvl.utils.caching.memcached import MemcachedBackend

def test_without_pymemcache(mocker):
    mocker.patch.dict('sys.modules', {'pymemcache': None})
    with pytest.raises(ModuleNotFoundError):
        backend = MemcachedBackend()

```

mdbenito (commented):
Left a couple of comments. But my main concern is about an in-memory cache that won't be exploited at all with the default config since worker processes only compute one marginal. Except for semivalues that can batch, but this is not default behaviour and it isn't documented that one should do use it.

src/pydvl/utils/caching/config.py
AnesBenmerzoug: I understand and agree with your point. But I would rather leave it as is or even set it back to `0.3` for now because I still feel that caching may not actually be useful (We can determine that once we work on #459).

I will create a separate issue for your suggestion of determining the time_threshold dynamically.

AnesBenmerzoug:
@mdbenito Thanks for the review! I addressed the code related comments.

As for the `InMemoryCacheBackend` class, I think you're wrong when you say that it isn't useful because it is used for only one marginal computation. Except for the semivalues module which submits one marginal computation at a time, all others (e.g. least core, permutation and combinatorial monte carlo) make multiple calls to the utility in a single worker so it is useful for those cases.

I thought about using what you suggested above, [inter-process shared memory](https://docs.python.org/3.8/library/multiprocessing.shared_memory.html) or, easier, a [multiprocessing manager](https://docs.python.org/3/library/multiprocessing.html#managers), but I wasn't sure about investing more time in that before making sure that caching is actually useful (#459).

If you still think that the current implementation of `InMemoryCacheBackend` is not useful, I can remove it.

src/pydvl/utils/caching/config.py
mdbenito: I believe 0 is not an option: it will be a performance hit and potentially create a bad network bottleneck. So 0.3 (although I think the dynamic thing is trivial to implement). Also, caching is necessary for at least two reasons:
* PermutationSampler
* Utility reuse for benchmarks (to avoid variance in the utility function).

AnesBenmerzoug: I will set it back to `0.3` for now. I agree that a simple implementation of the dynamic time_threshold estimation would be trivial but I would rather do it separately since I'm not convinced of the usefulness of caching, at least the way we are using it right now.

I would suggest thinking about utility reuse (e.g. for benchmarking) on its own and not try to tack it on caching.

Perhaps we could think of having an actual storage backend for the utility values by adding it directly to a Utility subclass (e.g. `StorageUtility`) or by replacing caching with key-value stores by an actual database as it is done in Optuna.

mdbenito:
This is currently not working 

```python
u1 = Utility(model, data, scorer)
u2 = Utility(model2, data, scorer)

v1 = compute_values(u1)  # misses cache
v2 = compute_values(u2)  # misses cache
v3 = compute_values(u1)  # hits cache
```

schroedk (approved):

PR #498, opened 2024-02-12
Add parameter precompute_grad to CgInfluence init, adapt documentation

### LINKED ISSUE #497: Memory leak in CgInfluence
Working with a Resnet50 modek on imagenet revealed a memory issue with the implementation of [CgInfluence](https://github.com/aai-institute/pyDVL/blob/develop/src/pydvl/influence/torch/influence_function_model.py#L425).

The issue is due to precomputing the averaged gradient over the training dataset, while keeping the computation graph for computing the second order derivatives. This is not feasible for large datasets.

**Resolution**:

Expose the parameter `precopmute_grad`. Setting to False, results in the averaging of the second order derivatives over the batches, which requires less memory (no need to keep the computation graph) but is slower in general.

 The same applies to [ArnoldiInfluence](https://github.com/aai-institute/pyDVL/blob/develop/src/pydvl/influence/torch/influence_function_model.py#L717).

schroedk:
### Description

This PR closes #497

### Changes

- Expose the parameter `precopmute_grad`. Setting to False, results in the averaging of the second order derivatives over the batches, which requires less memory (no need to keep the computation graph) but is slower in general.

### Checklist

- [ ] ~~Wrote Unit tests (if necessary)~~
- [x] Updated Changelog
- [ ] ~~If notebooks were added/changed, added boilerplate cells are tagged with `"tags": ["hide"]` or `"tags": ["hide-input"]`~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 
 - Bug in `LissaInfluence`, when not using CPU device [PR #495](https://github.com/aai-institute/pyDVL/pull/495)
+- Memory issue with `CgInfluence` and `ArnoldiInfluence`[PR #498](https://github.com/aai-institute/pyDVL/pull/498)
 
 ## 0.8.1 - 🆕 🏗  New method and noteboo, Games with exact shapley values, bug fixes and cleanup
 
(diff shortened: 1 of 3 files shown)


Xuzzo (approved):
Looks good to me! Thanks for the fix

PR #511, opened 2024-03-11
Fix pydvl.bib file by export from zotero, fix references in documenta…

### LINKED ISSUE #508: Inconsistent bibtex file for references
The file `docs/assets/pydvl.bib` has miss-matching keys to the Zotero collection. The style BibTex vs. BibLatex
is inconsistent within the file.

This should be harmonized and the export style from Zotero should be mentioned in the `CONTRIBUTING.md` guide.

schroedk:
### Description

This PR closes #508 

### Changes

- fix pydvl.bib file by export from zotero, fix references in documentation

### Checklist

- [x] ~Updated Documentation (if necessary)~
- [x] ~Updated Changelog~


mdbenito (changes requested):
I think this is good, but we lost the urls to the papers. Confusingly, mkdocs is automagically constructing links to arxiv, or using dois. But we should use the url in Zotero instead. If this means using the other bibtex format, that's ok. Compatibility with the website was a bogus argument since the yaml export is entirely independent.

PR #520, opened 2024-03-18
MSR method for Banzhaf

### LINKED ISSUE #519: Implement MSR for Banzhaf
As detailed in Sec 4.3 of [the paper](https://proceedings.mlr.press/v206/wang23e/wang23e.pdf).
See the open branch.

jakobkruse1:
### Description

This PR closes #519 

### Changes

- Implementing the MSR Banzhaf data valuation method

### Checklist

- [x] Write code for MSR Banzhaf method
- [x] Write notebook that showcases MSR Banzhaf and compares it with normal Banzhaf
- [x] Ensure that the method works in parallel computation (Marginal has memory now which may be an issue)
- [x] Wrote Unit tests
- [x] Updated Documentation
- [x] Updated Changelog


jakobkruse1:
@mdbenito @AnesBenmerzoug Any idea about the CI error? I don't think this is my fault. Have you seen that one before?

mdbenito:
> @mdbenito @AnesBenmerzoug Any idea about the CI error? I don't think this is my fault. Have you seen that one before?

It's this: https://github.com/squidfunk/mkdocs-material/issues/6983

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 - New method: `NystroemSketchInfluence`
   [PR #504](https://github.com/aai-institute/pyDVL/pull/504)
+- New method: `MSR Banzhaf Semivalues`
+  [PR #504](https://github.com/aai-institute/pyDVL/pull/520)
```
AnesBenmerzoug, line 10:
```suggestion
  [PR #520](https://github.com/aai-institute/pyDVL/pull/520)
```

docs/value/semi-values.md
```diff
@@ -98,6 +98,33 @@ values = compute_banzhaf_semivalues(
...
+a specific data point was included in the data subset or not. 
+The mathematical formula for computing the semivalues is
+
+$$\hat{\phi}_{MSR}(i) = \frac{1}{|\mathbf{S}_{\ni i}|} \sum_{S \in \mathbf{S}_{\ni i}} U(S)
+        - \frac{1}{|\mathbf{S}_{\not{\ni} i}|} \sum_{S \in \mathbf{S}_{\not{\ni} i}} U(S)$$
+
+where $\mathbf{S}_{\ni i}$ are the subsets that contain the index $i$ and $\mathbf{S}_{\not{\ni} i}$ are the 
+subsets not containing the index $i$.
+
+The function implementing this mehod is
```
AnesBenmerzoug, line 114:
```suggestion
The function implementing this method is
```

docs/value/semi-values.md
```diff
@@ -98,6 +98,33 @@ values = compute_banzhaf_semivalues(
...
+[compute_msr_banzhaf_semivalues][pydvl.value.semivalues.compute_msr_banzhaf_semivalues].
+```python
+from pydvl.value import *
```
AnesBenmerzoug, line 117: I think it's better to import exactly what we need for this snippet:

```python
from pydvl.value import compute_msr_banzhaf_semivalues, RankStability
```

P.S. There should be other snippets that use star imports and we should fix those as well in another PR.

src/pydvl/value/stopping.py
```diff
@@ -623,3 +625,74 @@ def reset(self):
...
+    r"""A check for stability of Spearman correlation between checks.
+
+    When the change in rank correlation between two successive iterations is
+    below a given threshold, the computation is terminated.
+    The criterion computes the Spearman correlation between two successive iterations.
+    The Spearman correlation uses the ordering indices of the given values and correlates them.
+    This means it focuses on the order of the elements instead of their exact values.
+    If the order stops changing (meaning the Banzhaf semivalues estimates converge), the criterion stops the algorithm.
+
+    This criterion is used in :footcite:t:`wang_data_2022`.
```
AnesBenmerzoug, line 640: `footcite` doesn't work with mkdocs. Look at [oob.py](https://github.com/aai-institute/pyDVL/blob/develop/src/pydvl/value/oob/oob.py) for an example of how we do it currently.

src/pydvl/value/sampler.py
```diff
@@ -312,6 +312,26 @@ def weight(cls, n: int, subset_len: int) -> float:
         return float(2 ** (n - 1)) if n > 0 else 1.0
 
 
+class MSRSampler(StochasticSamplerMixin, PowersetSampler[IndexT]):
+    """An iterator to perform sampling of random subsets.
+
+    This sampler does not return any index, it only returns subsets of the data.
+    This sampler is used in :footcite:t:`wang_data_2022`.
```
AnesBenmerzoug, line 319: `footcite` doesn't work with mkdocs. Look at [oob.py](https://github.com/aai-institute/pyDVL/blob/develop/src/pydvl/value/oob/oob.py) for an example of how we do it currently.

tests/value/test_semivalues.py
```diff
@@ -52,15 +54,29 @@ def test_marginal_batch_size(test_game, sampler, coefficient, batch_size, seed):
...
+@pytest.mark.parametrize("num_samples", [5])
+def test_msr_banzhaf(
+    num_samples: int, analytic_banzhaf, parallel_config, n_jobs, seed: Seed
+):
+    u, exact_values = analytic_banzhaf
+    values = compute_msr_banzhaf_semivalues(
+        u=u, done=MaxChecks(1000), config=parallel_config, n_jobs=n_jobs, seed=seed
+    )
+    # Need to use atol because msr banzhaf is quite noisy.
+    check_values(values, exact_values, atol=0.1)
```
AnesBenmerzoug, line 77: **Suggestion** Perhaps it would make sense to all make sure that the rankings are the same.

src/pydvl/value/sampler.py
```diff
@@ -44,6 +44,9 @@
       Frank, and Geoffrey Holmes. [Sampling Permutations for Shapley Value
       Estimation](http://jmlr.org/papers/v23/21-0439.html). Journal of Machine
       Learning Research 23, no. 43 (2022): 1–46.
+[^2]: <a name="wang2023data"></a>Wang, J.T. and Jia, R., 2023.
```
mdbenito, line 47: This citekey seems to be wrong. It should be `wang_2023_data`. If it's not like that in the bibtex file in `docs/assets/pydvl.bib`, change it there too.

---

We use Zotero for bibliography and betterbiblatex to export with the citekey format
```
[auth:lower:postfix,_][veryshorttitle:skipwords:lower:postfix,_][year]
```
Alas this is the convention from the tfl website and is not detailed [in the docs](https://github.com/aai-institute/pyDVL/blob/develop/CONTRIBUTING.md#using-bibliography). For this PR however you only need to change the id in the key that's already present in that file

notebooks/msr_banzhaf_spotify.ipynb
mdbenito, line 1: It seems like this notebook is mostly a copy of the spotify notebook. We could instead just add Banzhaf to that one. But I think it makes more sense to look at the purported benefits of banzhaf instead of repeating the experiments. It is supposed to help with stochastic utilities, so it should perform best when the model is stochastic, like a neural network. GBTs are however deterministic in nature.

I suggest to consider a different problem and create a new notebook for it in a separate PR. For this one, we could take the MSR sampling experiments and add them to the documentation for Banzhaf. Although first I would change the plot to have a different x-axis. Instead of iterations, which are not comparable if I'm not mistaken, I would take "number of utility evaluations"

src/pydvl/value/stopping.py
```diff
@@ -623,3 +628,74 @@ def reset(self):
 
     def __str__(self):
         return f"HistoryDeviation(n_steps={self.n_steps}, rtol={self.rtol})"
+
+
+class RankStability(StoppingCriterion):
```
mdbenito, line 633: I would add a `fraction` parameter ∈ [0,1] to this, to allow looking only at some of the data. Ususally one wants to focus on the top or bottom segment only. For the bottom, maybe allow a negative value for `fraction`?

mdbenito (commented):
Good stuff. I left some suggestions

docs/value/semi-values.md
```diff
@@ -98,6 +98,33 @@ values = compute_banzhaf_semivalues(
 )
 ```
 
+### Banzhaf semi-values with MSR sampling
```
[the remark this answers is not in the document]
jakobkruse1: Why would you do that :D Is there any way to properly set this in any IDE on this planet? @mdbenito

src/pydvl/value/stopping.py
jakobkruse1: Are you talking about only looking at the highest value fraction of the data or just the first data in the dataset? Sorting the data values, then looking at the highest value fraction does not make sense for checking the order of the values (because if would always be ordered and stop in Iteration 1). How exactly do you envision this feature to work?

notebooks/msr_banzhaf_spotify.ipynb
jakobkruse1: You are right, I should use another dataset and remove some of the copied snippets and add some new ones. The idea is to cover both Banzhaf values and also MSR sampling in one notebook. Number of iterations works, I can change that.

docs/value/semi-values.md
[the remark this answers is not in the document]
AnesBenmerzoug: In PyCharm, you can go to Settings -> Editor -> Code Styles and set a a hard wrap and visual guides at those specific widths. 

You could also create a `.editorconfig` file for this specific project (perhaps we can add it directly to the repository) with the following example content:

```toml
[*]
charset = utf-8
end_of_line = lf
indent_size = 4
indent_style = space
ij_visual_guides = 80,88
ij_wrap_on_typing = true

[{*.markdown,*.md}]
max_line_length = 80

[{*.py,*.pyw}]
max_line_length = 88
```

mkdocs.yml
```diff
@@ -140,6 +141,7 @@ plugins:
       type: iso_date
       fallback_to_build_date: true
   - social:
+      enabled: false
```
AnesBenmerzoug, line 144: If I'm not mistaken, this shouldn't be needed anymore after merging develop into this branch. Right, @mdbenito ?

src/pydvl/value/stopping.py
mdbenito: I meant "check stability among top k% of the data". So, for two runs, take the top k% points and compute ... 🤦🏽 as I'm typing this I realize I was thinking of what we did in the repro: look at the % of points that remain across the top k% positions across runs. Come to think of it, how is Spearman correlation a good stopping criterion? It's always going to be dismal across the whole dataset, isn't it?

mkdocs.yml
mdbenito: Correct

src/pydvl/value/stopping.py
jakobkruse1: This criterion would make sense, however, it does not involve the Spearman correlation at all and thus would be an entirely new Stopping criterion. I suggest to open a separate PR for this.

mdbenito:
@jakobkruse1 Please remove all warnings and irrelevant outputs (e.g. supress warnings, clean progress bars) before committing the notebook.

PS: I assigned myself to the PR. Our convention is assignee == merger, and merger is an additional reviewer.

jakobkruse1:
> @jakobkruse1 Please remove all warnings and irrelevant outputs (e.g. supress warnings, clean progress bars) before committing the notebook.
> 
> PS: I assigned myself to the PR. Our convention is assignee == merger, and merger is an additional reviewer.

Still working on finishing the notebook. Should be done soonish.
I can not suppress the warnings because they are raised in the parallel backend and suppressing them does not work as usual there. Any hint on how to do this?

jakobkruse1:
Notebook is running, I will push everything once it is done. Then we can merge today.

mdbenito (dismissed):

mdbenito (approved):

Issue #514, opened 2024-03-18
Move list of papers to documentation


AnesBenmerzoug:
Move the paper list to the documentation and link to the implementation as well as to the paperpill whenever it exists (which it should).

Issue #482, opened 2023-12-28
Timestamps in documentation pages


mdbenito:
"Last update" and "created" time displayed at the bottom of documentation pages always coincide. 
![image](https://github.com/aai-institute/pyDVL/assets/1548327/e4df53dc-9a0b-4124-bd11-5964c7a7ecb7)

Issue #462, opened 2023-12-03
Documentation improvements


AnesBenmerzoug:
After conducting user interviews, we realized that the current documentation has some issues and that those should be resolved because it is the the most used source of information for users.

We could have a look at the [lakefs-spec](https://lakefs-spec.org/latest/) documentation and use parts of it as inspiration.

# General

- [x] #514 
- [x] Make it easier to find the examples.
- [ ] #534
- [x] #524
- [x] #525

# First Steps

- [ ] #517
- [ ] #535
- [ ] #536

# Example

- [x] Show binder link more prominently.

AnesBenmerzoug:
Closing this issue in favour of smaller ones.

PR #521, opened 2024-03-18
Documentation fix and cleanup

### LINKED ISSUE #515: Use absolute URLs for images in order for them to render properly on pypi.

AnesBenmerzoug:
### Description

This PR closes #515 #514 #482  and part of #462 

### Changes

- Changed fetch depth in CI to 0 for jobs that build documentation.
- Used absolute URLs for images in readme.
- Moved list of papers from readme to a new page in documentation.
- Added new abbreviations.
- Added missing data banzhaf reference to bibliography.
- Improved mkdocs hook for modifying binder links in rendered notebooks.
- Added beautifulsoup4 as a docs dependency.
- Bumped version of some docs requirements.

### Checklist

- [ ] Updated Changelog

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -101,6 +101,8 @@ jobs:
       group: publish
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
       - name: Setup Python 3.8
         uses: ./.github/actions/python
         with:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 
 - Bump versions of CI actions to avoid warnings [PR #502](https://github.com/aai-institute/pyDVL/pull/502)
 - Add Python Version 3.11 to supported versions [PR #510](https://github.com/aai-institute/pyDVL/pull/510)
+- Documentation improvements and cleanup [PR #521](https://github.com/aai-institute/pyDVL/pull/521)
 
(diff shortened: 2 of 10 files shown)


.github/workflows/main.yaml
```diff
@@ -101,6 +101,8 @@ jobs:
       group: publish
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
schroedk, line 105: Just out of curiosity, why do we have to add this?

schroedk (approved):

.github/workflows/main.yaml
AnesBenmerzoug: The default value is set to `1` (See https://github.com/actions/checkout) and that is fine for most jobs except for the one that builds the docs because we use an mkdocs plugin that sets the creation and last update times in the lower left corner of the documentation pages based on git history. If it cannot determine the correct dates then it just uses the current date as shown in the image below.

image

AnesBenmerzoug:
@schroedk should I also add specific instructions for installing the influences extra requirements in this PR or will you take care of it?

.github/workflows/main.yaml
schroedk: Thanks for the explanation:)

schroedk:
> @schroedk should I also add specific instructions for installing the influences extra requirements in this PR or will you take care of it?

I will take care of it

PR #543, opened 2024-03-25
Fix/lfs images for docs

### LINKED ISSUE #482: Timestamps in documentation pages

### LINKED ISSUE #540: Fix rendering of influences plot in readme
With the recent change in #532 that moved all png images to git lfs, the influences plot in the readme no longer renders properly.

image

**EDIT** This could be related to the recent change we made to use absolute urls in the readme.

mdbenito:
### This PR:

- Closes #540 
- Closes #482
- Moves examples images to `examples/img`
- Fixes a broken link
- Adds svgs to lfs

### Checklist

- [x] Updated Changelog


.github/workflows/main.yaml
```diff
@@ -57,6 +57,7 @@ jobs:
     - uses: actions/checkout@v4
       with:
         fetch-depth: 0
+        lfs: true
```
AnesBenmerzoug, line 60: I think this should also be set in the publish workflow.

AnesBenmerzoug (dismissed):

.github/workflows/main.yaml
mdbenito: Thx good catch

.github/workflows/publish.yaml
```diff
@@ -29,11 +29,14 @@ jobs:
       - name: Checking out last commit in release
         if: ${{ github.event_name != 'workflow_dispatch' }}
         uses: actions/checkout@v4
+        with:
```
AnesBenmerzoug, line 32: I just noticed that we also need to set `fetch-depth: 0` here in order to make the timestamps work in the documentation.

mdbenito: So that closes #482 ? If you have already tested it, please push to this branch and merge! 🙏🏽

AnesBenmerzoug: Yes, If you check the development version of the docs, you can see that the timestamps were fixed there. I just forgot to also update the publish workflow.

mdbenito: Ok, I'll add it then

PR #549, opened 2024-03-27
Simplify parallel backend config

### LINKED ISSUE #385: Simplify parallel backend configuration
Now that we added a `joblib` backend and have switched to using joblib inside `MapReduceJob`, the parallel backend classes have become nothing more than simple wrappers (i.e. none of their methods, except `effective_n_jobs` are used). We could simply get rid of it and ask users to `ray.init` themselves if they want to configure ray and use the `joblib` context managers to configure the different joblib backend.

However the Futures executor interface still relies on `put` and `get` (e.g. in [`truncated_montecarlo_shapley`](https://github.com/appliedAI-Initiative/pyDVL/blob/develop/src/pydvl/value/shapley/truncated.py#L193)) and we should find a solution to replace that first.

AnesBenmerzoug:
### Description

This PR closes #385

### Changes

- Deprecated `config` argument of parallel backend classes, ray future executor and `MapReduceJob` class.
- Allowed instantiating parallel backend classes directly and deprecated `init_parallel_backend` function.
- Updated docstrings.
- Adapted all methods to use `parallel_backend` and deprecate `config` argument.

### Checklist

- [x] Updated Changelog

### DIFF
--- a/docs/getting-started/advanced-usage.md
+++ b/docs/getting-started/advanced-usage.md
@@ -48,18 +48,23 @@ to compute exact shapley values you would use:
 
 ```python
 import joblib
-from pydvl.parallel import ParallelConfig
+from pydvl.parallel import JoblibParallelBackend
 from pydvl.value.shapley import combinatorial_exact_shapley
 from pydvl.utils.utility import Utility
 
-config = ParallelConfig(backend="joblib") 
+parallel_backend = JoblibParallelBackend() 
 u = Utility(...)
 
 with joblib.parallel_config(backend="loky", verbose=100):
-    combinatorial_exact_shapley(u, config=config)
+    values = combinatorial_exact_shapley(u, parallel_backend=parallel_backend)
 ```
 
 #### Ray
+
+!!! info
+   
+    The Ray parallel backend requires optional dependencies.
+    See [Extras][installation-extras] for more information.
 
 Please follow the instructions in Ray's documentation to
 [set up a remote cluster](https://docs.ray.io/en/latest/cluster/key-concepts.html).
@@ -90,14 +95,14 @@ To use the ray parallel backend to compute exact shapley values you would use:
 
 ```python
 import ray
-from pydvl.parallel import ParallelConfig
+from pydvl.parallel import RayParallelBackend
 from pydvl.value.shapley import combinatorial_exact_shapley
 from pydvl.utils.utility import Utility
 
 ray.init()
-config = ParallelConfig(backend="ray")
+parallel_backend = RayParallelBackend()
 u = Utility(...)
-combinatorial_exact_shapley(u, config=config)
+vaues = combinatorial_exact_shapley(u, parallel_backend=parallel_backend)
 ```
 
 ### Influence functions
--- a/docs/getting-started/advanced-usage.md
+++ b/docs/getting-started/advanced-usage.md
@@ -16,7 +16,7 @@ keep in mind when using pyDVL namely Parallelization and Caching.
 pyDVL uses parallelization to scale and speed up computations. It does so
 using one of Dask, Ray or Joblib. The first is used in
 the [influence][pydvl.influence] package whereas the other two
-are used in the [value][pydvl.value] package. 
+are used in the [value][pydvl.value] package.
 
 ### Data valuation
 
@@ -36,6 +36,24 @@ and to provide a running cluster (or run ray in local mode).
     will typically make a copy of the whole model and dataset to each worker, even
     if the re-training only happens on a subset of the data. This means that you
     should make sure that each worker has enough memory to handle the whole dataset.
+
+We use backend classes for both joblib and ray as well as two types
+of executors for the different algorithms: the first uses a map reduce pattern as seen in
+the [MapReduceJob][pydvl.parallel.map_reduce.MapReduceJob] class
+and the second implements the futures executor interface from [concurrent.futures][].
+
+!!! info
+
+    The executor classes are not meant to be instantiated and used by users
+    of pyDVL. They are used internally as part of the computations of the 
+    different methods.
+
+!!! info
+
+    We are currently planning to deprecate
+    [MapReduceJob][pydvl.parallel.map_reduce.MapReduceJob] in favour of the
+    futures executor interface because it allows for more diverse computation
+    patterns with interruptions.
 
 #### Joblib
 
@@ -103,6 +121,50 @@ ray.init()
 parallel_backend = RayParallelBackend()
 u = Utility(...)
 vaues = combinatorial_exact_shapley(u, parallel_backend=parallel_backend)
+```
+
+#### Futures executor
+
+For the futures executor interface, we have implemented an executor 
+class for ray in [RayExecutor][pydvl.parallel.futures.ray.RayExecutor]
+and rely on joblib's loky [get_reusable_executor][loky.get_reusable_executor]
+function to instantiate an executor for local parallelization.
+
+They are both compatibles with the builtin
+[ThreadPoolExecutor][concurrent.futures.ThreadPoolExecutor]
+and [ProcessPoolExecutor][concurrent.futures.ProcessPoolExecutor]
+classes.
+
+```pycon
+>>> from joblib.externals.loky import _ReusablePoolExecutor
+>>> from pydvl.parallel import JoblibParallelBackend
+>>> parallel_backend = JoblibParallelBackend() 
+>>> with parallel_backend.executor() as executor:
+...     results = list(executor.map(lambda x: x + 1, range(3)))
+...
+>>> results
+[1, 2, 3]
+```
+
+#### Map reduce
+
+The map reduce interface is older and more limited in the patterns
+it allows us to use.
+
+To reproduce the previous example using
+[MapReduceJob][pydvl.parallel.map_reduce.MapReduceJob]we would use:
+
+```pycon
+>>> from pydvl.parallel import JoblibParallelBackend, MapReduceJob
+>>> parallel_backend = JoblibParallelBackend() 
+>>> map_reduce_job = MapReduceJob(
+...     list(range(3)),
+...     map_func=lambda x: x[0] + 1,
+...     parallel_backend=parallel_backend,
+... )
+>>> results = map_reduce_job()
+>>> results
+[1, 2, 3]
 ```
 
 ### Influence functions
--- a/docs/getting-started/advanced-usage.md
+++ b/docs/getting-started/advanced-usage.md
@@ -57,7 +57,7 @@ parallel_backend = init_parallel_backend(backend_name="joblib")
     of pyDVL. They are used internally as part of the computations of the 
     different methods.
 
-!!! info
+!!! danger "Deprecation notice"
 
     We are currently planning to deprecate
     [MapReduceJob][pydvl.parallel.map_reduce.MapReduceJob] in favour of the
@@ -88,7 +88,7 @@ with joblib.parallel_config(backend="loky", verbose=100):
 
 #### Ray
 
-!!! info
+!!! warning "Additional dependencies"
    
     The Ray parallel backend requires optional dependencies.
     See [Extras][installation-extras] for more information.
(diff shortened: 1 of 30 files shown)


src/pydvl/parallel/map_reduce.py
```diff
@@ -125,7 +139,17 @@ def __call__(
...
-        with Parallel() as parallel:
+
+        if isinstance(self.parallel_backend, JoblibParallelBackend):
+            backend = "loky"
+        elif isinstance(self.parallel_backend, RayParallelBackend):
+            backend = "ray"
+        else:
+            raise ValueError(
+                f"Unexpected parallel backend {self.parallel_backend.__class__.__name__}"
+            )
```
mdbenito, line 150: I'm sure you don't like this either. :) But this makes user code unable to add parallel backends and use mapreduce with them. Not likely, but still, we could add a property to ParallelBackend to avoid this ifelse. Or add a third object to map backends to backends, but that seems overkill. Especially since joblib only also supports dask...

docs/getting-started/advanced-usage.md
mdbenito, line 1: I think we need additional documentation on using `MapReduce`, and using the executors. In particular it is important to explain that objects must be `put()` and what it means. Also, the respective module's documentations need examples of low-level usage

mdbenito (commented):
As usual, good stuff. I left some low-value/interest comments, and one about documentation which I do think is important.

src/pydvl/parallel/backends/joblib.py
```diff
@@ -19,50 +20,72 @@
...
+        cancel_futures: Union[CancellationPolicy, bool] = CancellationPolicy.NONE,
     ) -> Executor:
+        """Returns a futures executor for the parallel backend.
+
+        !!! Example
+            ``` python
+            from pydvl.parallel import JoblibParallelBackend
+            parallel_backend = JoblibParallelBackend()
+            with parallel_backend.executor() as executor:
+                executor.submit(...)
```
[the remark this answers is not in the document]
AnesBenmerzoug: Your suggestion is exactly what we had before this PR with `init_parallel_backend`. I think the registry pattern hurts discoverability and readability and it's better to be more explicit in this case especially since changing this is considered "advanced usage" and the user should know what they're doing.

src/pydvl/parallel/backends/joblib.py
```diff
@@ -19,50 +20,72 @@
...
+                "and will be removed in v0.10.0.",
+                FutureWarning,
+            )
+
         if cancel_futures not in (CancellationPolicy.NONE, False):
             warnings.warn(
-                "Cancellation of futures is not supported by the joblib backend"
+                "Cancellation of futures is not supported by the joblib backend",
             )
-        return cast(Executor, get_reusable_executor(max_workers=max_workers))
```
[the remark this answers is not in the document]
AnesBenmerzoug: I think that cast should only be used sporadically because it's less readable than the alternative, at least for me. I will put it back the way it was for the sake of consistency though.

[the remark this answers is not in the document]
mdbenito: Quite the opposite, I find cast rather more readable than the alternative. I had to look twice at those lines because of the unexpected extra variable, whereas cast clearly conveys the intention.

src/pydvl/parallel/backends/joblib.py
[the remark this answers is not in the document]
mdbenito: No it's not "exactly what we had before"! I don't suggest going back, just a convenience function in case I want a factory to use based e.g. on whatever comes out of `os.getenv()`. But if you hate it, I guess people can just write an if/else

kosmitive:
Really cool! I would also vote for making that whole parallel stuff a separate library it is really useful as a standalone package.

src/pydvl/parallel/backend.py
```diff
@@ -41,25 +43,26 @@ class CancellationPolicy(Flag):
...
+        ParallelBackend.BACKENDS[backend_name] = cls
 
     @classmethod
     @abstractmethod
     def executor(
         cls,
-        max_workers: int | None = None,
-        config: ParallelConfig = ParallelConfig(),
-        cancel_futures: CancellationPolicy = CancellationPolicy.PENDING,
+        max_workers: Optional[int] = None,
```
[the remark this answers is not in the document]
AnesBenmerzoug: I personally prefer the use of `Optional[<type>]` over `<type> | None`. But I will put this back the way it was before.

docs/getting-started/advanced-usage.md
AnesBenmerzoug: `put()` is actually no longer used inside `MapReduceJob`. It is only used in the executors.

src/pydvl/parallel/backends/joblib.py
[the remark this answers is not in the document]
AnesBenmerzoug: I changed the deprecation for `init_parallel_backend` to only the config argument and allowed passing the backend name as a string.

src/pydvl/parallel/backends/joblib.py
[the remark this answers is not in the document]
AnesBenmerzoug: I have changed it back.

tests/parallel/test_parallel.py
```diff
@@ -116,14 +107,16 @@ def test_map_reduce_job(map_reduce_job_and_parameters, indices, expected):
...
+def test_chunkification(parallel_backend, data, n_chunks, expected_chunks):
+    map_reduce_job = MapReduceJob(
+        [], map_func=lambda x: x, parallel_backend=parallel_backend
+    )
     chunks = list(map_reduce_job._chunkify(data, n_chunks))
     for x, y in zip(chunks, expected_chunks):
         assert np.all(x == y)
 
 
-def test_map_reduce_job_partial_map_and_reduce_func(parallel_config):
```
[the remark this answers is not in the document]
AnesBenmerzoug: I don't think we need that for pyDVL especially since we're not planning to make this into a full blown parallelization library.

src/pydvl/parallel/backends/ray.py
```diff
@@ -17,27 +18,25 @@
...
 
-    def __init__(self, config: ParallelConfig):
+    @deprecated(
+        target=True,
+        args_mapping={"config": None},
+        deprecated_in="0.9.0",
+        remove_in="0.10.0",
+    )
+    def __init__(self, config: Optional[ParallelConfig] = None) -> None:
         if not ray.is_initialized():
```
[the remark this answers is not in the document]
AnesBenmerzoug: I did not understand what you said. Could you please give me more details?

src/pydvl/parallel/map_reduce.py
AnesBenmerzoug: For now, I changed this to raise an error if the passed parallel backend is not an instance of `JoblibParallelBackend`.

mdbenito: If I understand it correctly, option 1 is a no-go. As you know, we have non-map-reduce jobs. Plus we need to signal running processes and so on. As to option 2, I'm not sure I follow. I would either leave this as it is, or add an attribute  `ParallelBackend.joblib_backend`. Both hacks, but the latter is at least local to the backends themselves

tests/parallel/test_parallel.py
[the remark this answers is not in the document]
mdbenito: I think it's better to remain within one parallelization framework at a time. For the subtasks, ray recommends [nested tasks](https://docs.ray.io/en/latest/ray-core/patterns/nested-tasks.html).

src/pydvl/parallel/backends/ray.py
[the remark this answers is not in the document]
kosmitive: 1. The `ray.init` call has to be done somewhere. We can always say it is part of the user codebase, but we could also give a function `init_backend` for testing purposes.

2. `runtime_env` is defined in https://docs.ray.io/en/latest/ray-core/api/doc/ray.runtime_env.RuntimeEnv.html

tests/parallel/test_parallel.py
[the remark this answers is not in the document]
kosmitive: Sounds reasonable.

docs/getting-started/advanced-usage.md
mdbenito: I know. This doesn't make the docs less necessary.

src/pydvl/parallel/backends/joblib.py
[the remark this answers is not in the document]
AnesBenmerzoug: I have added a small snippet to the documentation about it.

docs/getting-started/advanced-usage.md
```diff
@@ -37,6 +37,33 @@ and to provide a running cluster (or run ray in local mode).
...
+parallel_backend = init_parallel_backend(backend_name="joblib")
+```
+
+!!! info
+
+    The executor classes are not meant to be instantiated and used by users
+    of pyDVL. They are used internally as part of the computations of the 
+    different methods.
+
+!!! info
```
mdbenito, line 60:
```suggestion
!!! danger "Deprecation notice"
```

docs/getting-started/advanced-usage.md
```diff
@@ -48,19 +75,24 @@ to compute exact shapley values you would use:
...
 u = Utility(...)
 
 with joblib.parallel_config(backend="loky", verbose=100):
-    combinatorial_exact_shapley(u, config=config)
+    values = combinatorial_exact_shapley(u, parallel_backend=parallel_backend)
 ```
 
 #### Ray
 
+!!! info
```
mdbenito, line 91:
```suggestion
!!! warning "Additional dependencies"
```

docs/getting-started/advanced-usage.md
```diff
@@ -90,14 +122,58 @@ To use the ray parallel backend to compute exact shapley values you would use:
...
+>>> from pydvl.parallel import JoblibParallelBackend
+>>> parallel_backend = JoblibParallelBackend() 
+>>> with parallel_backend.executor() as executor:
+...     results = list(executor.map(lambda x: x + 1, range(3)))
+...
+>>> results
+[1, 2, 3]
+```
+
+#### Map reduce
```
mdbenito, line 158:
```suggestion
#### Map-reduce
```

docs/getting-started/advanced-usage.md
```diff
@@ -90,14 +122,58 @@ To use the ray parallel backend to compute exact shapley values you would use:
...
+
+The map reduce interface is older and more limited in the patterns
```
mdbenito, line 160:
```suggestion
The map-reduce interface is older and more limited in the patterns
```

docs/getting-started/advanced-usage.md
```diff
@@ -90,14 +122,58 @@ To use the ray parallel backend to compute exact shapley values you would use:
...
+it allows us to use.
+
+To reproduce the previous example using
+[MapReduceJob][pydvl.parallel.map_reduce.MapReduceJob]we would use:
```
mdbenito, line 164:
```suggestion
[MapReduceJob][pydvl.parallel.map_reduce.MapReduceJob], we would use:
```

docs/getting-started/advanced-usage.md
AnesBenmerzoug: Sorry, I didn't mean it that way :sweat_smile:  I meant to say that I didn't know what to write. I have created issue #555
for that.

src/pydvl/parallel/__init__.py
```diff
@@ -1,37 +1,37 @@
...
-from pydvl.parallel import init_executor, ParallelConfig
+from pydvl.parallel import JoblibParallelBackend
 
-config = ParallelConfig(backend="joblib")
-with init_executor(config=config) as executor:
+parallel_backend = JoblibParallelBackend()
+with parallel_backend.executor(max_workers=2) as executor:
     results = list(executor.map(lambda x: x + 1, range(5)))
 assert results == [1, 2, 3, 4, 5]
 ```
```
mdbenito, line 30:
```suggestion
!!! tip "Passsing large objects"
    When running tasks which accept heavy inputs, it is important
    to first use `put()` on the object and use the returned reference
    as argument to the callable within `submit()`. For example:    
    ```python
    u_ref = parallel_backend.put(u)
    ...
    executor.submit(task, utility=u)
    ```
    Note that `task()` does not need to be changed in any way:
    the backend will `get()` the object and pass it to the function
    upon invocation.

mdbenito (commented):
Left some changes. I will apply them and merge

docs/getting-started/advanced-usage.md
mdbenito: Partially addressed in https://github.com/aai-institute/pyDVL/pull/549/files/e23256ba2166fbb6cfd58eb1f15663702b0e1b0a#r1553250280

mdbenito (approved):
Good schtuff

PR #558, opened 2024-04-09
Feature/refactor value

### LINKED ISSUE #187: Include number of trials in tolerate fixture
So that instead of:

```python
@pytest.mark.parametrize("num_attempts, max_failures, some_arg", [(10, 1, foo), (10,2,bar)])
def test_blah(tolerate, num_attempts, max_failures, some_arg):
    for _ in range(num_attempts):
        with tolerate(max_failures):
            do_random_stuff(some_arg)
```

we can do:
```python
@pytest.mark.parametrize("num_attempts, max_failures, some_arg", [(10, 1, foo), (10,2,bar)])
def test_blah(tolerate, num_attempts, max_failures, some_arg):
        with tolerate(num_attempts, max_failures):
            do_random_stuff(some_arg)
```

or:

```python
@pytest.mark.tolerate(num_attempts=10, max_failures=1)
def test_blah():
       do_random_stuff()
```

### LINKED ISSUE #223: Implement stratified sampling
* [ ] Implement the optimal strategy described in [1].
* [x] #226
* [ ] Reproduce their results (possibly, but not necessarily including the stratified sampling strategy of [2]), in particular all information in Tables 1-5

**[1]** Castro, Javier, Daniel Gómez, Elisenda Molina, and Juan Tejada. “Improving Polynomial Estimation of the Shapley Value by Stratified Random Sampling with Optimum Allocation.” Computers & Operations Research 82 (June 1, 2017): 180–88. https://doi.org/10.1016/j.cor.2017.01.019.
**[2]** Maleki, Sasan, Long Tran-Thanh, Greg Hines, Talal Rahwan, and Alex Rogers. “Bounding the Estimation Error of Sampling-Based Shapley Value Approximation.” ArXiv:1306.4265 [Cs], February 12, 2014. http://arxiv.org/abs/1306.4265.

### LINKED ISSUE #295: Consistency in argument names
The argument name n_iterations in least core and group testing doesn't really make much sense. It's the number of constraints used in the problem. Before, it made sense to use n_iterations everywhere for consistency, but now the main MC methods use StoppingCriterion. And since LC and GT need a dedicated parameter, which in both cases is a number of constraints or samples to take, maybe we should change the name to be n_samples or n_constraints in both.

We have to agree first on what to use and then document it before making changes.

### LINKED ISSUE #321: Interruptible samplers
#319 introduces `PermutationSampler` but it does not include the possibility of interrupting the sampling within a permutation, as required for TCMS. 

One possibility would be to make samplers not simple `Iterables` but coroutines, with `__iter__` returning a `Generator[NDArray, bool, None]` that accepts sent booleans to interrupt the sampling of a permutation. With this (and given that caching is enabled, see the comment in `semivalues.py`, but this is unrelated to the interruption), `permutation_montecarlo_shapley` and `semivalues` with `shapley_coefficient` and `PermutationSampler` should be equivalent.

Additionally, stratified samplers might require either simple interruption, or information from the utility computations. For instance, an adaptive variance-reducing sampler might require the status (running moments?) for each stratum separately.

### LINKED ISSUE #325: Subsume shapley values into semi-value computation
#319 duplicates some of the methods in `value.shapley`. We will want to merge as much as possible of the latter into the semi-values, possibly just by calling `semivalues()` from within the methods in `value.shapley`.

### LINKED ISSUE #350: Use scorer's default value for all utility computations
We don't always use the default value of the Scorer.  For instance when iterating over a permutation, we might want to do:

```python
prev_score = u({})
```

Then we also have a similar hardcoded choice in [Utility._utility()](https://github.com/appliedAI-Initiative/pyDVL/blob/42d492d3eed9f90e13357ba57e03c7ed963b534c/src/pydvl/utils/utility.py#L176), where we always return 0.0 for an empty index set, and to be consistent, we should return `self.scorer.default`. However, this breaks user code, because the previous default value for `scorer.default` was `np.nan`, and not 0.0.

We could remove the default assignment and force the user to select one.

As to the choice of 0.0, we should switch to the score of a constant classifier / regressor (using the mean of the data).

_Originally posted by @mdbenito in https://github.com/appliedAI-Initiative/pyDVL/issues/346#issuecomment-1528779970_

### LINKED ISSUE #490: Issue with SemiValue batching and parallelization
While working on PR #341, I realized that there is a bug in the batching feature of semivalues when using `n_jobs` > 1.
The results are almost the same but not exactly the same.

### LINKED ISSUE #512: Deprecate compute_*_values interface
In order to prepare for a newer interface for game theoretic data valuation methods i.e. #467, we should deprecate the [`compute_shapley_values`](https://github.com/aai-institute/pyDVL/blob/96326ae75600391e927b01bf2e31a93f12915159/src/pydvl/value/shapley/common.py) and [`compute_least_core_values`](https://github.com/aai-institute/pyDVL/blob/96326ae75600391e927b01bf2e31a93f12915159/src/pydvl/value/least_core/__init__.py#L42) functions.

### LINKED ISSUE #539: Move tolerate fixture to separate repository
With the changes in #529, we no longer use the tolerate fixture and we should therefore remove it from the repository. However I think it is useful and there may be other uses for it beyond pyDVL. Therefore I suggest we move it a separate repository and open-source it.

All the related code can be found in [tests/tolerate.py](https://github.com/aai-institute/pyDVL/blob/develop/tests/tolerate.py) and [tests/conftest.py](https://github.com/aai-institute/pyDVL/blob/develop/tests/conftest.py).

We should refer to [pytest's documentation](https://docs.pytest.org/en/7.1.x/how-to/writing_plugins.html#writing-your-own-plugin) for instructions on how to create a plugin.

It turns out that there is a similar plugin called [pytest-repeat](https://github.com/pytest-dev/pytest-repeat) that allows repeating tests on failure. However our implementation allows setting the number of repetitions inside a test's function. We could contribute that change to that other plugin but that would take longer and would involve a lot more communication overhead.

### LINKED ISSUE #613: Potential bug in KNN Shapley
## Description of the algorithm

The KNN Shapley method is based on this [paper](https://arxiv.org/pdf/1908.08619)

The basic algorithm is: 

image

The shapley values are calculated independently for each test data point and then averaged to the overall shapley values. 

## Our implementation

In [our implementation](https://github.com/aai-institute/pyDVL/blob/217afb65a85fb9fcc0a555f33d69011a53645fdb/src/pydvl/value/shapley/knn.py#L76), there is a dependence between the shapley values of each test data point:

```python
values: NDArray[np.float_] = np.zeros_like(u.data.indices, dtype=np.float_)
n = len(u.data)
yt = u.data.y_train
iterator = enumerate(zip(u.data.y_test, indices), start=1)
for j, (y, ii) in tqdm(iterator, disable=not progress):
    value_at_x = int(yt[ii[-1]] == y) / n
    values[ii[-1]] += (value_at_x - values[ii[-1]]) / j
    for i in range(n - 2, n_neighbors, -1):  # farthest to closest
        value_at_x = (
            values[ii[i + 1]] + (int(yt[ii[i]] == y) - int(yt[ii[i + 1]] == y)) / i
        )
        values[ii[i]] += (value_at_x - values[ii[i]]) / j
    for i in range(n_neighbors, -1, -1):  # farthest to closest
        value_at_x = (
            values[ii[i + 1]]
            + (int(yt[ii[i]] == y) - int(yt[ii[i + 1]] == y)) / n_neighbors
        )
        values[ii[i]] += (value_at_x - values[ii[i]]) / j
```

The relevant part is that `values` is initialized outside the loop over data points but then used to initialize the recursive calculation of shapley values inside the loop in each iteration. Starting with the second iteration, this produces different results. 

## Test results 

The differences in implementation seem to matter for the results. I tested a new implementation that is very close to the paper against montecarlo estimates and the old implementation. The new implementation agrees with the montecarlo estimates but the old one does not. Here are the result:

```python
# montecarlo estimates
array([0.06966294, 0.06785734, 0.06547832, 0.06693986, 0.06346014,  0.06554825, 0.06678741])

# new results
array([0.06573427, 0.06573427, 0.06713287, 0.06713287, 0.06713287,  0.06713287, 0.06573427])

# old results
array([0.16653024, 0.13782045, 0.16464412, 0.14736292, 0.19502404,  0.08048527, 0.13009609])
```

mdbenito:
### Description

This PR closes #XXX

closes #539
closes #187 
closes #613 
closes #223 
closes #325
closes #321
closes #490
closes #350
closes #295 
closes #512
### Changes

- 

### Checklist

- [x] Updated Changelog


src/pydvl/value/games.py
```diff
@@ -92,13 +92,13 @@ class DummyModel(SupervisedModel):
     def __init__(self) -> None:
         pass
 
-    def fit(self, x: NDArray, y: NDArray) -> None:
```
AnesBenmerzoug, line 95: Shouldn't this module moved to the new valuation package?

P.S. This comment is not about this specific line. I just found it easier to comment on it.

src/pydvl/valuation/methods/semivalue.py
```diff
@@ -0,0 +1,142 @@
...
+    def fit(self, data: Dataset):
+        self.result = ValuationResult.zeros(
+            # TODO: automate str representation for all Valuations (and find something better)
+            algorithm=f"{self.__class__.__name__}-{self.utility.__class__.__name__}-{self.sampler.__class__.__name__}-{self.is_done}",
+            indices=data.indices,
+            data_names=data.data_names,
+        )
+        ensure_backend_has_generator_return()
+
+        self.utility.training_data = data
```
AnesBenmerzoug, line 105:
```suggestion
        self.utility.with_dataset(data)
```

src/pydvl/valuation/methods/semivalue.py
```diff
@@ -0,0 +1,142 @@
...
+    def coefficient(self, n: int, k: int) -> float:
+        """Computes the coefficient for a given subset size.
+
+        Args:
+            n: Total number of elements in the set.
+            k: Size of the subset for which the coefficient is being computed
+        """
+        ...
+
+    def fit(self, data: Dataset):
```
AnesBenmerzoug, line 96: Why not change the signature to this to allow using the dataset set on the utility, if there is one of course?

```suggestion
    def fit(self, data: Dataset | None = None):
```

src/pydvl/valuation/utility/modelutility.py
```diff
@@ -0,0 +1,239 @@
...
+
+    """
+
+    model: ModelT
+    scorer: Scorer
+
+    def __init__(
+        self,
+        model: ModelT,
+        scorer: Scorer,
```
AnesBenmerzoug, line 115: The scorer is no longer the model's default scorer. Was it changed on purpose?

src/pydvl/valuation/methods/beta_shapley.py
```diff
@@ -0,0 +1,36 @@
...
+class BetaShapleyValuation(SemivalueValuation):
+    """Computes Beta-Shapley values."""
+
+    algorithm_name = "Beta-Shapley"
+
+    def __init__(
+        self,
+        utility: UtilityBase,
+        sampler: IndexSampler,
+        is_done: StoppingCriterion,
```
janosg, line 21: This comes out of a discussion with @schroedk: `is_done` sounds like a bool. We should discuss renaming this across all methods to something like `stopping` or `stop_criterion`

schroedk (approved):

PR #565, opened 2024-04-19
Add `with_dataset` method to Utility to avoid side effect

janosg:


janosg:
I have a better way of doing this and will open another PR for it.

PR #574, opened 2024-05-03
Add functionality to set a device fixture depending on the availability

### LINKED ISSUE #573: Add the possibility to manually test on cuda devices
To test the device handling of our torch related code, we should add the possibility to manually run tests with cuda
support.

Suggestion:

- Add `--with-cuda` option:

  ```python
  parser.addoption(
          "--with-cuda",
          action="store_true",
          default=False,
          help="Use CUDA for testing if available",
      )
  ```
- Add a device fixture, which can be used in tests:
  ```python
  @pytest.fixture(scope="session")
  def device(request):
      import torch
  
      use_cuda = request.config.getoption("--with-cuda")
      if use_cuda and torch.cuda.is_available():
          return torch.device("cuda")
      else:
          return torch.device("cpu")
  ``` 
- manually run `pytest --with-cuda`

schroedk:
### Description

This PR closes #573 

### Changes

- Add a device fixture, which depending on the availability and user input (--with-cuda) resolves to cuda device.

### Checklist

- [ ] Updated Changelog

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -131,6 +131,13 @@ There are a few important arguments:
 - `--slow-tests` enables running slow tests. See below for a description
   of slow tests.
 
+- `--with-cuda` sets the device fixture in [tests/influence/torch/conftest.py](
+  tests/influence/torch/conftest.py) to `cuda` if it is available.
+  Using this fixture within tests, you can run parts of your tests on a `cuda` 
+  device. Be aware, that you still have to take care of the usage of the device
+  manually in a specific test. Setting this flag does not result in
+  running all tests on a GPU.
+
 ### Markers
 
 We use a few different markers to differentiate between tests and runs
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -48,6 +48,12 @@ def pytest_addoption(parser):
         default=False,
         help="Disable reporting. Verbose mode takes precedence.",
     )
+    parser.addoption(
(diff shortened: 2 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -131,6 +131,13 @@ There are a few important arguments:
 - `--slow-tests` enables running slow tests. See below for a description
   of slow tests.
 
+- `--with-cuda` sets the device fixture in [tests/influence/torch/conftest.py](
+  tests/influence/torch/conftest.py) to `cuda` if it is available.
+  Using this fixture within tests, you can run parts of your tests on a `cuda` 
+  device. Be aware, that you still have to take care of the usage of the device
```
mdbenito, line 137: Maybe this can be a bit more explicit, eg say something like "Note that you need to use the fixture in your tests in order to benefit from this (...)"

mdbenito (approved):
Useful! But wouldn't one want to run tests both with and without cuda as well?

schroedk:
> Useful! But wouldn't one want to run tests both with and without cuda as well?

Yes, but I would ask the user to do it manually, and not automatically. In our automatic test environment, we do not have access to cuda anyway.

CONTRIBUTING.md
schroedk: For now, I would leave it like that, until we get feedback, that this is not understandable.

PR #578, opened 2024-05-06
Fix/552 ekfac bug mps osx

### LINKED ISSUE #552: EkfacInfluence breaks for large LinearLayer
Due to an [issue in pytorch](https://github.com/pytorch/pytorch/issues/92141), EkfacInfluence does not work for linear layers with more than **2895** input or output size.
This is caused by an issue in `torch::linalg::eigh` which leads to some 32bit integer overflow. The only fix is to use 64bit, pytorch is working on that AFAIK, see the issue [here](https://github.com/pytorch/pytorch/issues/51720).

In EkfacInfluence, we are using
```
evals_a, evecs_a = torch.linalg.eigh(forward_x[key])
evals_g, evecs_g = torch.linalg.eigh(grad_y[key])
```
with forward_x and grad_y coming from _init_layer_kfac_blocks:
```
sG = module.out_features
sA = module.in_features + int(with_bias)
forward_x_layer = torch.zeros((sA, sA), device=module.weight.device)
grad_y_layer = torch.zeros((sG, sG), device=module.weight.device)
```
Basically, the matrix `torch.zeros((sA, sA), ...)` is just too big to work with `torch.linalg.eigh`. This then raises a super ugly error message:
```
RuntimeError: false INTERNAL ASSERT FAILED at ".../aten/src/ATen/native/BatchLinearAlgebra.cpp":1462, 
please report a bug to PyTorch. linalg.eigh: Argument 8 has illegal value. 
Most certainly there is a bug in the implementation calling the backend library.
```

**Proposed fix**:
My proposed fix for EkfacInfluence is to check the sG and sA variables and make sure they are small enough not to run into the 32bit integer problem. This also affected the `influence_sentiment_analysis` notebook, but I already took the large linear layers out of Ekfac in another PR #550 .

schroedk:
### Description

This PR closes #552 

### Changes

- add a decorator to catch exceptions of a specific type and raise a more concrete exception type
- add `TorchLinalgEighException`
- add `safe_torch_linalg_eigh` which is decorated to raise the above exception, if a `RunTimeError` occurs in calling torch.linalg.eigh  
- add tests, where the exception testing is only pulled out on osx-arm64

### Checklist

- [x] Updated Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@
   implementation [PR #570](https://github.com/aai-institute/pyDVL/pull/570)
 - Missing move to device of `preconditioner` in `CgInfluence` implementation
   [PR #572](https://github.com/aai-institute/pyDVL/pull/572)
+- Raise a more specific error message, when a `RunTimeError` occurs in 
+  `torch.linalg.eigh`, so the user can check if it is related to a known
+  issue
+  [PR #578](https://github.com/aai-institute/pyDVL/pull/578)
 
 ## 0.9.1 - Bug fixes, logging improvement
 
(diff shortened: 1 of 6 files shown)


janosg (dismissed):
Very nice!

PR #580, opened 2024-05-08
Refactor/least core

janosg:
### Description

This PR refactors the least-core to make use of the new samplers in `pydvl.valuation`. The previously supported least-core methods are now just special cases: "exact" is obtained using a `DeterministicUniformSampler`; "montecarlo" is obtained using a `UniformSampler`. 

### Changes

Complete rewrite of the least-core code in `pydvl.valuation`. The old code remains unchanged in `pydvl.value`

### Checklist

- [x] Updated Changelog


src/pydvl/valuation/methods/_least_core_solving.py
```diff
@@ -0,0 +1,373 @@
...
+        algorithm=algorithm,
+        status=status,
+        values=values,
+        subsidy=subsidy,
+        stderr=None,
+        data_names=u.training_data.data_names,
+    )
+
+
+@deprecated(
```
mdbenito, line 160: If I understand correctly, this deprecation notice shouldn't be necessary:
1. Because the deprecation of the config argument is completed in 0.10.0 which is the release which will include the refactor
2. Because lc_solve_problems in this module is not exported (and distinct from the original one in `pydvl.value`

src/pydvl/valuation/samplers/base.py
```diff
@@ -8,7 +8,7 @@
 
 import logging
 from abc import ABC, abstractmethod
-from typing import Callable, Generic, TypeVar
+from typing import Callable, Generic, Optional, TypeVar
```
mdbenito, line 11: Note that we import `annotations` from `__future__` to have the pipe operator for types (sort of)

tests/valuation/__init__.py
mdbenito, line 1: When copying files you can use https://github.com/opcode81/git-split to keep the history. If you already made changes, I would: compute a diff to this from the original file, delete this one, split the original file to keep history, apply the patch. It's a bit of extra work, but it's nicer.

src/pydvl/reporting/scores.py
```diff
@@ -51,3 +54,49 @@ def compute_removal_score(
         score = u(indices)
         scores[pct] = score
     return scores
+
+
+def compute_removal_score_new(
```
mdbenito, line 59: As discussed, I'd simply move this function to the notebook support, where it can be adapted as necessary. The old one can be left here and marked for deprecation

mdbenito (commented):
I haven't really reviewed, but left a couple of comments for now

tests/valuation/__init__.py
janosg: I tried your suggestion but it did not produce what I expected. Here is the [produced history](https://github.com/aai-institute/pyDVL/commits/1fcc3cf64af6e5026e7142f454ca814299edc2b1/src/pydvl/valuation/methods/_least_core_solving.py).

src/pydvl/valuation/methods/_solve_least_core_problems.py
```diff
@@ -0,0 +1,336 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class LeastCoreProblem(NamedTuple):
+    utility_values: NDArray[np.float_]
+    A_lb: NDArray[np.float_]
+
+
+def lc_solve_problem(
```
schroedk, line 34: This function is mixing two levels of abstraction, on the one hand it is doing a lot of low level stuff to construct the conditions for the optimization problem and on the other hand, it calls other more low-level functions (_solve_...). This mixing of levels of abstractions makes it harder to read. Even if a user would not call this function directly, for the developers it is a more difficult task to grasp what is happening (actually the comment helps a lot and in combination with the log messages it seems to me that the related code should be encapsulated in other low-level functions with only one purpose).

src/pydvl/valuation/samplers/base.py
```diff
@@ -113,14 +112,33 @@ def from_data(self, data: Dataset) -> BatchGenerator:
...
         self._interrupted = False
+        self._n_samples = 0
         for batch in take_n(self._generate(indices), self.batch_size):
             yield batch
             # FIXME, BUG: this could be wrong if the batch is not full. Just use lists
             self._n_samples += self.batch_size
             if self._interrupted:
                 break
 
+    def length(self, indices: IndexSetT) -> int | None:
```
schroedk, line 130: Given what is happening here, maybe sample_limit could be a better name?

src/pydvl/valuation/samplers/base.py
```diff
@@ -113,14 +112,33 @@ def from_data(self, data: Dataset) -> BatchGenerator:
 
     def from_indices(self, indices: IndexSetT) -> BatchGenerator:
         """Batches the samples and yields them."""
+
```
schroedk, line 115: If I see something like object.from_something(something), I would expect that another instance of the object's type is returned. But actually this generates batches, so my first impulse would be to call it generate_batches_from_indices(indices) and generate_batches(data). This is a discussion, we often have, but I am in the group "Clarity over brevity"

src/pydvl/valuation/samplers/utils.py
```diff
@@ -10,7 +10,7 @@
 T = TypeVar("T")
 
 
-def take_n(it: Iterable[T], size: int = 2) -> Generator[Iterator[T], Any, None]:
+def take_n(it: Iterable[T], size: int = 2) -> Generator[List[T], Any, None]:
```
schroedk, line 13: There is a nice little package called more-itertools, see: https://more-itertools.readthedocs.io/en/stable/_modules/more_itertools/more.html#chunked

Just mentioning, in case you did not know it before. The name "take_n" is a bit missleading, as I would expect to get the first "n" (or "size") elements.  Actually, the definition of batch in line 23 is called "take" in more-itertools: https://more-itertools.readthedocs.io/en/stable/_modules/more_itertools/recipes.html#take

schroedk (changes requested):

src/pydvl/valuation/methods/least_core.py
```diff
@@ -1,28 +1,207 @@
...
+        u_values.extend(v)
+
+    utility_values = np.array(u_values)
+    A_lb = np.row_stack(masks).astype(float)
+
+    return LeastCoreProblem(utility_values=utility_values, A_lb=A_lb)
+
+
+def _correct_n_samples(candidate: int | None, sampler_length: int | None) -> int:
+    """Correct a user provided n_samples parameter
```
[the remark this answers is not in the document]
janosg: After adding subclasses for the special cases (ExactLeastCoreValuation and MonteCarloLeastCoreValuation) we don't need this correction anymore. Instead we only need to set a default sample size if no sample size was provided by the user.

src/pydvl/valuation/methods/least_core.py
```diff
@@ -1,28 +1,207 @@
...
+
 __all__ = ["LeastCoreValuation"]
 
 
 class LeastCoreValuation(Valuation):
+    """Umbrella class to calculate Least Core values with multiple sampling methods.
+
+    See [Data valuation][data-valuation] for an overview.
+
+    Different samplers correspond to different least-core methods from the literature:
```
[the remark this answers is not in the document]
janosg: I agree. Done.

src/pydvl/valuation/methods/least_core.py
```diff
@@ -1,28 +1,207 @@
...
+        _check_sampler(sampler)
+        self._utility = utility
+        self._sampler = sampler
+        self._non_negative_subsidy = non_negative_subsidy
+        self._solver_options = solver_options
+        self._n_samples = n_samples
+        self._progress = progress
+
+    def fit(self, data: Dataset) -> Valuation:
+        self._utility = self._utility.with_dataset(data)
```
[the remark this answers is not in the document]
janosg: I agree, done. Also added the NotFittedException if methods are not called in the correct order.

src/pydvl/valuation/methods/least_core.py
```diff
@@ -1,28 +1,207 @@
...
+def create_least_core_problem(
+    u: UtilityBase, sampler: PowersetSampler, n_samples: int, progress: bool
+) -> LeastCoreProblem:
+    """Create a Least Core problem from a utility and a sampler.
+
+    Args:
+        u: Utility object with model, data and scoring function.
+        sampler: The sampler to use for the valuation.
+        n_samples: The maximum number of samples to use for the valuation.
+        progress (bool): Whether to show a progress bar during the construction of the
```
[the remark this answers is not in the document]
janosg: Thank, done!

src/pydvl/valuation/methods/least_core.py
```diff
@@ -1,28 +1,207 @@
...
+            u_values.append(u(sample))
+
+        return masks, u_values
+
+    generator = takewhile(
+        lambda _: sampler.n_samples < n_samples,
+        sampler.from_indices(u.training_data.indices),
+    )
+
+    generator_with_progress = tqdm(  # type: ignore
```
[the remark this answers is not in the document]
janosg: Thank, done!

src/pydvl/valuation/samplers/base.py
```diff
@@ -113,14 +112,33 @@ def from_data(self, data: Dataset) -> BatchGenerator:
...
+        # early return for empty indices. Necessary because some samplers use a
+        # while True loop to generate infinite samples.
+        if len(indices) == 0:
+            return
```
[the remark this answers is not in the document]
janosg: I added a comment to explain why that is not the case. The reason is that `generate_batches` is a generator function (because it has a yield statement later in its body) and inside generator functions `return` acts more like a `break` statement in a loop. It's described in [this post](https://stackoverflow.com/a/13243870) on stackoverflow.

The simplest example for this behavior is:

```python
>>> def f():
...     return
...     yield 

# f is a generator
>>> type(f())
generator

# the generator is empty
>>> list(f())
[]

The return type is not None
>>> f() is None:
False
```

src/pydvl/valuation/methods/_solve_least_core_problems.py
```diff
@@ -0,0 +1,336 @@
...
+    "lc_solve_problems",
+    "LeastCoreProblem",
+]
+
+logger = logging.getLogger(__name__)
+
+
+class LeastCoreProblem(NamedTuple):
+    utility_values: NDArray[np.float_]
+    A_lb: NDArray[np.float_]
```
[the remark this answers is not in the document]
janosg: I added a docstring with links.

src/pydvl/valuation/methods/least_core.py
```diff
@@ -1,28 +1,207 @@
 from __future__ import annotations
 
+import math
+import warnings
+from itertools import islice, takewhile
```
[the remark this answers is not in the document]
janosg: Thanks, done!

src/pydvl/valuation/samplers/base.py
janosg: I agree, done!

src/pydvl/valuation/samplers/base.py
janosg: As discussed, I removed `from_data` and renamed `from_indices` to `generate_batches`.

src/pydvl/valuation/methods/least_core.py
```diff
@@ -1,28 +1,207 @@
...
+    """
+    if u.training_data is not None:
+        n_indices = len(u.training_data.indices)
+    else:
+        raise ValueError("Utility object must have a training dataset.")
+
+    batch_size = sampler.batch_size
+    n_batches = math.ceil(n_samples / batch_size)
+
+    def _process(
```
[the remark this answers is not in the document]
janosg: I agree. Did both.

src/pydvl/valuation/methods/_solve_least_core_problems.py
```diff
@@ -0,0 +1,336 @@
...
+        algorithm=algorithm,
+        status=status,
+        values=values,
+        subsidy=subsidy,
+        stderr=None,
+        data_names=u.training_data.data_names,
+    )
+
+
+def lc_solve_problems(
```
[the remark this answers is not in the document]
janosg: I wonder if we can completely remove `lc_solve_problems`. 

The reason why `lc_solve_problems` existed is described [here](https://pydvl.org/stable/api/pydvl/value/least_core/). In short: The creation of the problems can already be parallelized, the solving cannot. Therefore, if one runs many least-core valuations, one might want to separate problem creation and solving. `lc_solve_problems` can then run the multiple solves in parallel. 

It seems to me like one could get the same benefit by simply calling `LeastCoreValuation.fit` in parallel and the benefit of doing so would be that users don't need detailed knowledge about the fact that LeastCoreValuation internally performs a problem creation and a problem solving step.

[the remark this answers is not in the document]
janosg: As discussed, I removed `lc_solve_problems`.

src/pydvl/valuation/methods/least_core.py
```diff
@@ -1,28 +1,207 @@
...
-        self.solver_options = solver_options
-
-    def fit(self, data: Dataset):
-        pass
+
+        _check_sampler(sampler)
+        self._utility = utility
+        self._sampler = sampler
+        self._non_negative_subsidy = non_negative_subsidy
+        self._solver_options = solver_options
```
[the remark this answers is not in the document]
janosg: As discussed, a proper solution is out of scope for this PR. I created an Issue (#583).

src/pydvl/valuation/methods/_solve_least_core_problems.py
janosg: I created an issue about this and related improvements (#584)

src/pydvl/valuation/samplers/utils.py
janosg: Yes, `more_itertools.chunked` does exactly what `take_n` did. I replaced it everywhere and removed `take_n`.

schroedk (approved):

PR #582, opened 2024-05-21
Feature/499 datainf

### LINKED ISSUE #499: Implement DataInf: Efficiently Estimating Data Influence in LoRA-tuned LLMs and Diffusion Models
Implement [DataInf: Efficiently Estimating Data Influence in LoRA-tuned LLMs and Diffusion Models](http://arxiv.org/abs/2310.00902)

Code from the paper: https://github.com/ykwon0407/DataInf

### LINKED ISSUE #585: Increase flexibility of influence computation (block aproximation, Gauss-Newton approximation)
For almost all of the current implementations, it is possible to apply them in a block approximative fashion. In addition,
it is possible to not solve inverse Hessian problems, but a different kind of curvature approximation, as the Gauss-Newton matrix.

This issue is meant to implement a flexible backend, which makes it possible to:
- compose block structures
- easily replace Hessian by Gauss-Newton computations

The API for the existing implementations of `InfluenceFunctionModel` should not change doing this.

A first realization should be done with #499.

Proposed generic abstractions:
- `GradientProvider` (computes per sampel gradients) 
- `Operator` (maps gradients to transformed gradients)
- `OperatorGradientComposition` (combine the first two, to get something like a bi-linear form, which acts on (transformed-)gradients, these are the composable blocks for block diagonal approximations )
- `BlockMapper` (represents an ordered collection of composable blocks of type `OperatorGradientComposition` )

schroedk:
### Description

This PR closes #499 closes #585

### Changes

- New method `InverseHarmonicMeanInfluence`, implementation for the paper 
  `DataInf: Efficiently Estimating Data Influence in LoRA-tuned LLMs and 
    Diffusion Models` #499 
- Add new backend implementations for influence computation
  to account for block-diagonal approximations #585 

### Checklist

- [x] Updated Changelog


src/pydvl/influence/torch/base.py
```diff
@@ -0,0 +1,710 @@
...
+    def _inner_product(self, left: torch.Tensor, right: torch.Tensor) -> torch.Tensor:
+        left_result = self.operator.apply(left)
+
+        if left_result.ndim == right.ndim and left.shape[-1] == right.shape[-1]:
+            return left_result @ right.T
+
+        return torch.einsum("ia,j...a->ij...", left_result, right)
+
+
+class DictBilinearForm(OperatorBilinearForm):
```
janosg, line 296: My feeling is that we could simplify this by using something like `tree_map` together with `OperatorBilinearForm`. See #592 for more information.

src/pydvl/influence/torch/base.py
```diff
@@ -0,0 +1,710 @@
...
+        if right is None:
+            right = left
+        if left.shape[0] <= right.shape[0]:
+            return self._inner_product(left, right)
+        return self._inner_product(right, left).T
+
+    def _inner_product(self, left: torch.Tensor, right: torch.Tensor) -> torch.Tensor:
+        left_result = self.operator.apply(left)
+
+        if left_result.ndim == right.ndim and left.shape[-1] == right.shape[-1]:
```
janosg, line 290: I would prefer an explicit `else` over a second return statement. I find it most readable if methods have only one return statement.

src/pydvl/influence/torch/util.py
```diff
@@ -598,3 +608,141 @@ def __init__(self, original_exception: RuntimeError):
...
+                    warnings.warn(
+                        f"The parameter {parameter_name} from the block "
+                        f"{block_name} is mark as not trainable in the model "
+                        f"and will be excluded from the computation."
+                    )
+            parameter_dict[block_name] = inner_ordered_dict
+
+        return parameter_dict
+
+    def build_from_block_mode(
```
janosg, line 685: I would have expected that `build_from_block_mode` creates a `block_structure` and then calls `build`. Are there specific reasons why would not work here?

src/pydvl/influence/torch/util.py
```diff
@@ -598,3 +608,141 @@ def __init__(self, original_exception: RuntimeError):
...
+            to parameters.
+        """
+        parameter_dict = {}
+
+        if block_mode is BlockMode.FULL:
+            inner_ordered_dict = {}
+            for k, v in self.model.named_parameters():
+                if v.requires_grad:
+                    inner_ordered_dict[k] = self._optional_detach(v)
+            parameter_dict[""] = inner_ordered_dict
```
janosg, line 709: I would prefer a more explicit name over an empty string here.

tests/influence/torch/test_batch_operation.py
```diff
@@ -0,0 +1,309 @@
...
+    hvp_autograd = hessian_op.apply(TorchBatch(x, y), vec)
+    hvp_autograd_dict = hessian_op.apply_to_dict(
+        TorchBatch(x, y), align_structure(params, vec)
+    )
+    hvp_autograd_dict_flat = flatten_dimensions(hvp_autograd_dict.values())
+
+    assert torch.allclose(hvp_autograd, h_analytical @ vec, rtol=tol)
+    assert torch.allclose(hvp_autograd_dict_flat, h_analytical @ vec, rtol=tol)
+
+    op_then_flat = flatten_dimensions(
```
janosg, line 45: This test is quite long and if one of the early asserts fails you don't even get to the second ones. Maybe we could split the assertions that make sure the order of operations does not matter into a separate test?

janosg (commented):
I think the new abstractions in the `types` module are really good and will also work well for JAX. The rest are minor comments.

src/pydvl/influence/torch/base.py
schroedk: Sounds very good, I would like to postpone this to a separate PR.

src/pydvl/influence/torch/util.py
schroedk: Will check, if this can be simplified and made more readable.

src/pydvl/influence/torch/util.py
schroedk: Empty string for as identifier for the full model is a pytorch specific.

tests/influence/torch/test_batch_operation.py
schroedk: Agree, will split the test to make it more readable and group tests by semantics.

src/pydvl/influence/torch/util.py
schroedk: Done

janosg (approved):

PR #591, opened 2024-06-03
Feature/586 block diagonal direct

### LINKED ISSUE #586: Adapt `DirectInfluence` to use block-approximation
Extend `DirectInfluence` to handle:

1. Block-diagonal approximations
2. Gauss-Newton approximation

schroedk:
### Description

This PR closes #586 

### Changes

- Add functionality for block-diagonal and Gauss-Newton approximation to `DirectInfluence`

### Checklist

- [x] Updated Changelog


schroedk:
Blocked by #582

docs/influence/index.md
```diff
@@ -249,6 +249,87 @@ not to corrupt the outcome too much, the parameter $\lambda$ should be as small
...
+))
+# custom block-diagonal structure
+infl_model = DirectInfluence(model, loss,
+                             regularization=0.1,
+                             block_structure=block_structure)
+infl_model = infl_model.fit(training_data_loader)
+```
+If you would like to apply a block-specific regularization, you can provide a
+dictionary with the block names as keys and the regularization values as values.
+In this case, the specification must be complete, i.e. every block must have
```
janosg, line 285: We could also fill everything for which no regularization was provided with a default value (probably 0)

src/pydvl/influence/torch/functional.py
```diff
@@ -424,13 +437,16 @@ def hessian(
         def flat_input_batch_loss_function(
             p: torch.Tensor, t_x: torch.Tensor, t_y: torch.Tensor
         ):
-            return blf(align_with_model(p, model), t_x, t_y)
+            return blf(align_structure(params, p), t_x, t_y)
```
janosg, line 440: Took me a while to understand that `blf` means `batch_loss_function`. I think `batch_loss` would be much clearer than `blf` and not too long (unless this is again a strong pytorch convention or so)

src/pydvl/influence/torch/functional.py
```diff
@@ -447,6 +463,57 @@ def flat_input_empirical_loss(p: torch.Tensor):
...
+
+    Returns:
+        The Gauss-Newton matrix.
+    """
+
+    per_sample_grads = create_per_sample_gradient_function(model, loss)
+
+    params = restrict_to
+    if params is None:
+        params = {k: p.detach() for k, p in model.named_parameters() if p.requires_grad}
```
janosg, line 496: Dict comprehensions like this appear very often. Should we have a function `get_detached_params(model)` instead?

src/pydvl/influence/torch/influence_function_model.py
```diff
@@ -351,109 +353,115 @@ def to(self, device: torch.device):
...
+        ```
+        Accordingly, if you choose a layer-wise or parameter-wise structure
+        (by providing `BlockMode.LAYER_WISE` or `BlockMode.PARAMETER_WISE` for
+        `block_structure`) the keys must be the layer names or parameter names,
+        respectively.
+
+        You can retrieve the block-wise influence information from the methods
+        with suffix `_by_block`. By default, `block_structure` is set to
+        `BlockMode.FULL` and in this case these methods will return a dictionary
+        with the empty string being the only key.
```
janosg, line 398: This text is very similar to the description in `docs/influence/index.md`. Could we link to that place instead of repeating to make maintenance easier?

src/pydvl/influence/torch/influence_function_model.py
```diff
@@ -351,109 +353,115 @@ def to(self, device: torch.device):
...
+        block_params: Dict[str, torch.nn.Parameter],
+        data: DataLoader,
+        regularization: Optional[float],
+    ) -> TorchOperatorGradientComposition:
+        gp = TorchGradientProvider(self.model, self.loss, restrict_to=block_params)
 
-        """
-        return super().influences(x_test, y_test, x, y, mode=mode)
+        if self.second_order_mode is SecondOrderMode.GAUSS_NEWTON:
+            mat = gauss_newton(self.model, self.loss, data, restrict_to=block_params)
```
janosg, line 455: More a question than a comment: Is it efficient to materialize the Gauss Newton matrix and then use a direct solver? I thought it would be more efficient (and numerically robust) to build up the solutions from the factors of the Gauss Newton matrix (e.g. something like QR decomposition on the factors, triangular solve on R and then multiply).

docs/influence/index.md
schroedk: This is the actual behavior, but the documentation is wrong, as it describes a special case.

src/pydvl/influence/torch/functional.py
schroedk: Agree, will check

src/pydvl/influence/torch/influence_function_model.py
schroedk: Agree, will check

src/pydvl/influence/torch/influence_function_model.py
janosg: I created Issue #594 for this

src/pydvl/influence/torch/functional.py
schroedk: Renamed to `batch_loss`, done.

src/pydvl/influence/torch/functional.py
schroedk: Add a util function `get_model_parameters`, done.

src/pydvl/influence/torch/influence_function_model.py
schroedk: Linked to documentation, done.

janosg (approved):

PR #597, opened 2024-06-12
Refactor owen

### LINKED ISSUE #187: Include number of trials in tolerate fixture

### LINKED ISSUE #539: Move tolerate fixture to separate repository

janosg:
### Description

This PR refactors the Owen Shapley values to use the new sampler architecture.

### Changes

- Refactor Owen Shapley values 
- Fix a bug in `pydvl.utils.numeric.random_subset`
- Use pytest-rerunfailures instead of our tolerate fixture (closes #539 and closes #187)

### Checklist

- [x] Updated Changelog


CHANGELOG.md
```diff
@@ -7,6 +7,13 @@
 - Refactoring of least-core data valuation methods with more supported sampling methods
   and consistent interface.
   [PR #580](https://github.com/aai-institute/pyDVL/pull/580)
+- Refactoring of owen shapley valuation with new sampler architecture
+  [PR #597](https://github.com/aai-institute/pyDVL/pull/597)
+
+### Fixed
+
+- Fix a bug in pydvl.utils.numeric.random_subset
```
schroedk, line 15: Do we have an issue for the bug? Otherwise, I would very shortly describe the reason for the bug here

src/pydvl/utils/numeric.py
```diff
@@ -96,7 +96,7 @@ def random_subset(
         The subset
     """
     rng = np.random.default_rng(seed)
-    selection = rng.uniform(size=len(s)) > q
```
schroedk, line 99: I guess this is the bug fix, correct? So you could add "Fix a bug in ..., due to incorrect comparison operator"

src/pydvl/valuation/methods/owen_shapley.py
```diff
@@ -1,7 +1,73 @@
...
 ## References
 
 [^1]: <a name="okhrati_multilinear_2021"></a>Okhrati, R., Lipani, A., 2021.
     [A Multilinear Sampling Algorithm to Estimate Shapley Values](https://ieeexplore.ieee.org/abstract/document/9412511).
     In: 2020 25th International Conference on Pattern Recognition (ICPR), pp. 7992–7999. IEEE.
 """
+
+
+class OwenShapleyValuation(SemivalueValuation):
+    """Umbrella class to calculate least-core values with Owen sampling schemes.
```
schroedk, line 19: You could add a link to the documentation here, describing what Owen Shapley values are. But I guess the documentation is outdated by now, correct? Do you want to update the documentation in this PR?

src/pydvl/valuation/samplers/powerset.py
```diff
@@ -369,6 +368,95 @@ def _generate(self, indices: IndexSetT) -> SampleGenerator:
                 yield Sample(idx, subset)
 
 
+class OwenSampler(StochasticSamplerMixin, PowersetSampler):
+    """A sampler for Owen shapley values.
+
```
schroedk, line 373: Do you think it makes sense to add a small mathematical expression about how the sampling works here?

src/pydvl/valuation/samplers/powerset.py
```diff
@@ -369,6 +368,95 @@ def _generate(self, indices: IndexSetT) -> SampleGenerator:
...
+    def weight(n: int, subset_len: int) -> float:
+        return 1.0
+
+    def sample_limit(self, indices: IndexSetT) -> int:
+        return len(indices) * self._n_samples_outer * self._n_samples_inner
+
+
+class AntitheticOwenSampler(OwenSampler):
+    """A sampler for antithetic Owen shapley values.
+
```
schroedk, line 420: Maybe you can add a small description about the difference to `OwenSampler`

src/pydvl/valuation/types.py
```diff
@@ -52,6 +53,16 @@ def __hash__(self):
         sha256_hash = hashlib.sha256(self.subset.tobytes()).hexdigest()
         return int(sha256_hash, base=16)
 
+    def with_idx_in_subset(self) -> Self:
```
schroedk, line 56: I think this method should get a small docstring, in particular mentioning the error raise in case the index is not set.

janosg: Agree, done.

src/pydvl/utils/numeric.py
janosg: Agree, done.

src/pydvl/valuation/methods/owen_shapley.py
janosg: We'll do the update as part of the re-structuring.

schroedk (approved):

PR #605, opened 2024-06-19
Refactor MSR Banzhaf valuation

janosg:
### Description

This PR implements Maximum Sample Re-use (MSR) Banzhaf valuation in the new architecture. The implementation deviates strongly from the previous implementation and fixes a bug in the variance estimation. 

The new implementation uses two `ValuationResult` instances to keep track of the positive and negative running means. After each update, those are combined into the final result object. The update counter of the combined result is set to the minimum of the two update counters. The variance of the combined result is set to the sum of variances (assuming independence).

### Open questions

- How should update counters for MSR be defined? I chose the minimum of the positive and negative update counters because this makes the counters comparable to the counters for normal semivalues. 
- To which value should we set the value estimates if either the positive or the negative mean had zero updates? The paper suggests zero but I think `nan` is safer. Note that if the update counter is defined as suggested above, this situation can be completely avoided by using `MinUpdates` as stopping criterion. 
- To which value should we set the variance estimates if either the positive or the negative mean had zero updates? I chose `inf`. 

### Checklist

- [x] Updated Changelog


janosg:
I clarified the documentation of `ValuationResult.variance`, but I still think it is a bit misleading that `ValuationResult.variances` is not just the square of `ValuationResult.stderr`. 

I think it would be clearer if we only expose the square root of the variances as `ValuationResult.stdev`; Then it's clear that the difference betweenn `stdev` and `stderr` must be a conceptual one. Also, most of the time standard deviations are more interpretable than variances.

src/pydvl/valuation/methods/msr_banzhaf.py
```diff
@@ -0,0 +1,186 @@
...
+
+    """
+    # set counts to the minimum of the two; This enables us to ensure via stopping
+    # criteria that both running means have a minimal number of updates
+    counts = np.minimum(pos_result.counts, neg_result.counts)
+
+    values = pos_result.values - neg_result.values
+    values[counts == 0] = np.nan
+
+    variances = pos_result.variances + neg_result.variances
```
janosg, line 174: Besides assuming independence, this was only valid under my wrong assumption that `variances` are sampling variances that are already divided by the `counts` of each running mean. 

I think in our current approach of working with sample variances and then dividing them by root-n in the `stderr` property, it is not possible to get valid standard errors for MSR Banzhaf.

jakobkruse1 (commented):
I just looked over the files, seems all good to me at first glance. I did not run any code or tests yet though, so no way to be sure for me. Also the CI is not running for this branch. Are we updating the notebooks for the new valuation structure as well or is this planned at a later stage? Looking at notebooks may help to verify that everything works as expected.
I am not aware of the current plans about notebooks etc. so feel free to ignore me if there are any decisions that I am unaware of.

src/pydvl/valuation/methods/msr_banzhaf.py
```diff
@@ -0,0 +1,201 @@
+"""
+This module implements the MSR-Banzhaf valuation method, as described in
+(Wang et. al.)<sup><a href="wang_data_2023">3</a></sup>.
```
schroedk, line 3:
```suggestion
(Wang et. al.)<sup><a href="wang_data_2023">1</a></sup>.
```

src/pydvl/valuation/methods/msr_banzhaf.py
```diff
@@ -0,0 +1,201 @@
+"""
```
schroedk, line 1:
```suggestion
r"""
```

src/pydvl/valuation/methods/msr_banzhaf.py
```diff
@@ -0,0 +1,201 @@
...
+
+    """
+
+    algorithm_name = "MSR-Banzhaf"
+
+    def __init__(
+        self,
+        utility: UtilityBase,
+        sampler: MSRSampler,
+        is_done: StoppingCriterion,
```
schroedk, line 59: Reading just the argument name, I would expect a boolean. What about `stop_criterion`?

src/pydvl/valuation/methods/msr_banzhaf.py
```diff
@@ -0,0 +1,201 @@
...
+        progress: bool = True,
+    ):
+        sampler = MSRSampler()
```
schroedk, line 62: That is quite unexpected, because the user has to provide a sampler, but that instance is never used. As no message is logged, this could be highly confusing. What is the reason to instantiate a `MSRSampler` here?

src/pydvl/valuation/methods/msr_banzhaf.py
```diff
@@ -0,0 +1,201 @@
...
+        super().__init__(
+            utility=utility,
+            sampler=sampler,
+            is_done=is_done,
+            progress=progress,
+        )
+
+    @staticmethod
+    def coefficient(n: int, k: int) -> float:
```
schroedk, line 71: As the method in the base class is not static, what is the reason to make it static here?

src/pydvl/valuation/methods/msr_banzhaf.py
```diff
@@ -0,0 +1,201 @@
...
+
+        ```python
+        from joblib import parallel_config
+
+        with parallel_config(n_jobs=4):
+            valuation.fit(data)
+        ```
+
+        """
+        self._pos_result = ValuationResult.zeros(
```
schroedk, line 91: My IDE is complaining, that this is defined outside of init. Is this used to check `is_fitted`?

src/pydvl/valuation/methods/msr_banzhaf.py
```diff
@@ -0,0 +1,201 @@
...
+            indices=data.indices,
+            data_names=data.data_names,
+            algorithm=self.algorithm_name,
+        )
+
+        self._neg_result = ValuationResult.zeros(
```
schroedk, line 97: It looks like this could be a local variable.

src/pydvl/valuation/methods/msr_banzhaf.py
```diff
@@ -0,0 +1,201 @@
...
+## References
+
+[^1]: <a name="wang_data_2023"></a>Wang, J.T. and Jia, R., 2023.
+    [Data Banzhaf: A Robust Data Valuation Framework for Machine Learning](https://proceedings.mlr.press/v206/wang23e.html).
+    In: Proceedings of The 26th International Conference on Artificial Intelligence and Statistics, pp. 6388-6421.
+
+"""
+from __future__ import annotations
+
+import warnings
```
schroedk, line 14: Unused import

src/pydvl/valuation/methods/msr_banzhaf.py
```diff
@@ -0,0 +1,201 @@
...
+                )
+                for batch in Progress(delayed_evals, self.is_done, **self.tqdm_args):
+                    for evaluation in batch:
+                        if evaluation.kind == ValueUpdateKind.POSITVE:
+                            self._pos_result.update(evaluation.idx, evaluation.update)
+                        elif evaluation.kind == ValueUpdateKind.NEGATIVE:
+                            self._neg_result.update(evaluation.idx, evaluation.update)
+                        else:
+                            raise ValueError(
+                                "Invalid ValueUpdateKind: {evaluation.kind}"
```
schroedk, line 130:
```suggestion
f"Invalid ValueUpdateKind: {evaluation.kind}"
```

src/pydvl/valuation/methods/msr_banzhaf.py
```diff
@@ -0,0 +1,201 @@
...
+    given our convention for the counts and the normal calculation of standard errors
+    in the valuation result.
+
+    Note that we cannot use the normal addition or subtraction defined by the
+    ValuationResult because it is weighted with counts. If we were to simply subtract
+    the negative result from the positive we would get wrong variance estimates,
+    misleading update counts and even wrong values if no further precaution is taken.
+
+    TODO: Verify that the two running means are statistically independent (which is
+    assumed in the aggregation of variances).
```
schroedk, line 170:
```suggestion
 assumed in the aggregation of variances).
```

src/pydvl/valuation/methods/msr_banzhaf.py
```diff
@@ -0,0 +1,201 @@
...
+                            self.sampler.interrupt()
+                            break
+
+                    if self.is_done(self.result):
+                        break
+
+        return self
+
+
+def _combine_results(
```
schroedk, line 148: As it is only used once in the class `MSRBanzhafValuation`, I would include it as a static method.

src/pydvl/valuation/types.py
```diff
@@ -29,10 +30,17 @@
 NullaryPredicate = Callable[[], bool]
 
 
+class ValueUpdateKind(str, Enum):
```
schroedk, line 33: Maybe a short docstring would be helpful, e.g. what does `DEFAULT` mean, is it context dependend?

src/pydvl/valuation/methods/msr_banzhaf.py
janosg: Very good catch. That was a leftover from when I thought there would only be one MSR sampler. But in the future we might add weighted ones. Removed it.

src/pydvl/valuation/methods/msr_banzhaf.py
janosg: Thanks!

src/pydvl/valuation/methods/msr_banzhaf.py
janosg: Agree, thanks!

src/pydvl/valuation/methods/msr_banzhaf.py
janosg: Thanks!

src/pydvl/valuation/methods/msr_banzhaf.py
janosg: Thanks!

src/pydvl/valuation/types.py
janosg: As discussed, we got rid of `ValueUpdateKind`

src/pydvl/valuation/methods/msr_banzhaf.py
janosg: Done.

src/pydvl/valuation/methods/msr_banzhaf.py
janosg: Made a comment on #558 as this is not specific to the current PR

src/pydvl/valuation/samplers/msr.py
```diff
@@ -0,0 +1,76 @@
...
+    ValueUpdate,
+    ValueUpdateKind,
+)
+from pydvl.valuation.utility.base import UtilityBase
+
+__all__ = ["MSRSampler"]
+
+
+class MSRSampler(StochasticSamplerMixin, IndexSampler):
+    def __init__(self, batch_size: int = 1, seed: Seed | None = None):
```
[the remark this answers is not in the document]
janosg: Yes, thanks!

src/pydvl/valuation/methods/msr_banzhaf.py
```diff
@@ -0,0 +1,196 @@
...
+        standard errors in the valuation result.
+
+        Note that we cannot use the normal addition or subtraction defined by the
+        ValuationResult because it is weighted with counts. If we were to simply
+        subtract the negative result from the positive we would get wrong variance
+        estimates, misleading update counts and even wrong values if no further
+        precaution is taken.
+
+        TODO: Verify that the two running means are statistically independent (which is
+        assumed in the aggregation of variances).
```
schroedk, line 165:
```suggestion
 assumed in the aggregation of variances).
```

schroedk (approved):

Issue #581, opened 2024-05-13
Restructure documentation


schroedk:
Restructure the documentation into:
- explanations
- tutorials
- how-to guides
- api

Take inspiration from
https://nick.groenen.me/posts/the-4-types-of-technical-documentation/

AnesBenmerzoug:
As another suggestion, we should write the how-to guides and tutorials as notebooks and execute them, when possible, in CI with caching (with jupyter-cache?) in order to make sure they always run successfully. We should also consider putting directly in the corresponding subdirectory of the docs directory to avoid having to copy them every time.

mdbenito:
Given the current lack of human-power I am moving this to v0.11.0. I will try to implement some of these ideas in #653

PR #609, opened 2024-07-02
Restructure parallelization and Caching documentation

AnesBenmerzoug:
### Description

This PR addresses part of #581 

### Changes

- Created a how-to guide for speeding up computations using parallelization using existing documentation and adapting it to changes in #558.
- Created a how-to guide for speeding up computations using caching using existing documentation and adapting it to changes in #558.

### Checklist

- [ ] ~Updated Changelog~


docs/how-to-guides/how-to-speedup-caching.md
```diff
@@ -0,0 +1,88 @@
...
+from pydvl.utils.caching.disk import DiskCacheBackend
+from pydvl.valuation.utility import ModelUtility
+
+cache_backend = DiskCacheBackend()
+utility = ModelUtility(..., cache_backend=cache_backend)
+```
+
+Please refer to the documentation and examples of each backend class for more details.
+
+!!! tip "When is the cache really necessary?"
```
mdbenito, line 57: We should mention that this method is deprecated with the new interface.

docs/how-to-guides/how-to-speedup-caching.md
```diff
@@ -0,0 +1,88 @@
...
+
+In general, caching won't play a major role in the computation of Shapley values
+because the probability of sampling the same subset twice, and hence needing
+the same utility function computation, is very low. However, it can be very
+useful when comparing methods that use the same utility function, or when
+running multiple experiments with the same data.
+
+pyDVL supports 3 different caching backends:
+
+- [InMemoryCacheBackend][pydvl.utils.caching.memory.InMemoryCacheBackend]:
```
mdbenito, line 26: We should add a note that this will be deprecated and changed to shared memory since only wrapping the model into routines that release the GIL can benefit from it.

mdbenito (commented):
Just a couple of comments. Haven't reviewed all of it.

A more general note is: does this really count as a "how to" guide? If I understood correctly, those should focus on answering the question, not on explaining how things work. Opinions anyone?

janosg:
> A more general note is: does this really count as a "how to" guide? If I understood correctly, those should focus on answering the question, not on explaining how things work. Opinions anyone?

I would also say that in the current form it is mixing a how-to guide and explanations. An ideal how-to guide according to the diataxis framework is extremely focused on solving one specific user need. Users are usually in a hurry and don't read anything carefully. So the how-to guide needs to be very easy to skim. Less text and more code snippets is better. But you do need a bit of text (headings and introductory sentences) to reassure the reader they are at the right place or tell them where they should go instead. 

I think I also have made the mistake of adding explanations in almost all how-to guides I wrote. But I'll try to make an example of how I understand how-to guides now. 

An example of a specific situation could be: A user wants to do data valuation. It's slow but they have a big computer and wonder how they can use it. 

Then the how-to guide could be structured as follows:

# How to speed up data valuation with parallelization

This guide will show you how to speed up data valuation algorithms by using parallel hardware. For alternative ways to speed up data valuation see [caching]. For parallelization of influence functions see [...].

## Parallelization on a single computer

This approach is a good idea if you are working on a laptop or small server. 

```python
# code snippet here
# the code should already make it obvious how one sets the number of cores
# no need to mention default backends or stuff
```

*(Explain what the code does but do not describe in words how it could be changed or how joblib works. If you want to show variations, add a second code snippet. If there are many code snippets, use tabs.)* 

For advanced configuration see [joblib docs]

## Parallelization on multiple machines 

...

AnesBenmerzoug:
Thanks for the feedback! I have restructured and updated the content.

I want your opinion on a few ideas:

- I think we should write the how-to guide as notebook to make sure they always run on CI and thus reduce the likelihood of them going stale.
- I think we should move the notebooks to the specific place where they're needed / used inside the docs directory. This would simplify the documentation build process because we can get rid of the build script that copies the notebooks everytime.

AnesBenmerzoug:
@janosg, @schroedk I made a few more changes and added more information. This should be ready for a final review.

docs/how-to-guides/how-to-scale-up-if-with-parallel.md
```diff
@@ -1,39 +1,55 @@
...
-to iterate over the batches sequentially and collect them. pyDVL provides the simple convenience class
-[SequentialInfluenceCalculator][pydvl.influence.influence_calculator] to do this. The
-batch size should be chosen as large as possible, such that the corresponding batches fit
-into memory.
+    For speeding-up data valuation algorithms with parallelization
+    see [[speed-up-value-with-parallel]].
+
+!!! warning "Speeding up Influence Function"
+
+    Scaling up Influence Function computations will not speed up
```
schroedk, line 22: Well, this is only true, if all data fits into memory and the additional overhead would not be needed. But this holds true in great generality, so I would not specifically mention it here.

docs/how-to-guides/how-to-scale-up-if-with-parallel.md
```diff
@@ -1,39 +1,55 @@
...
+    the computations and may even slow them down.
+
+### Sequential Computation
+
+The simplest way to compute influence functions is sequentially using the
```
schroedk, line 27:
```suggestion
The simplest way to compute influence functions is using the convenience class 
```

docs/how-to-guides/how-to-scale-up-if-with-parallel.md
```diff
@@ -1,39 +1,55 @@
...
-    TorchNumpyConverter, NestedTorchCatAggregator,
-)
 from pydvl.influence.torch import CgInfluence
+from torch.utils.data import DataLoader
 
 batch_size = 10
 train_dataloader = DataLoader(..., batch_size=batch_size)
 test_dataloader = DataLoader(..., batch_size=batch_size)
 
 infl_model = CgInfluence(model, loss, hessian_regularization=0.01)
```
schroedk, line 44:
```suggestion
infl_model = CgInfluence(model, loss, regularization=0.01)
```

docs/how-to-guides/how-to-scale-up-if-with-parallel.md
```diff
@@ -107,27 +128,45 @@ test_data_set: Dataset = LargeDataSet(
 train_dataloader = DataLoader(train_data_set)
 infl_model = CgInfluence(model, loss, hessian_regularization=0.01)
```
schroedk, line 129:
```suggestion
infl_model = CgInfluence(model, loss, regularization=0.01)
```

docs/how-to-guides/how-to-scale-up-if-with-parallel.md
AnesBenmerzoug: I removed it.

docs/how-to-guides/how-to-speedup-caching.md
AnesBenmerzoug: I removed it.

docs/how-to-guides/how-to-speedup-caching.md
AnesBenmerzoug: I added a warning next to it.

docs/how-to-guides/how-to-speed-up-value-with-parallel.md
```diff
@@ -0,0 +1,197 @@
...
+utility = ModelUtility(model, scorer)
+training_set, test_set = Dataset.from_sklearn(data, random_state=16)
+valuation = DataShapleyValuation(
+    utility,
+    sampler=PermutationSampler(batch_size=10, seed=16),
+    is_done=MaxUpdates(100)
+)
+```
+
+## Parallelization
```
janosg, line 71: This section is very good but it is explaining background instead of showing how to get stuff done. I would keep it in this document but switch the order of "Local Parallelization" and "Parallelization" and rename "Parallelization" to something like "Understanding the pattern" or "Understanding what happened". That way people first see the code example they can copy and paste and then can decide on their own if they want to continue reading.

docs/how-to-guides/how-to-scale-up-if-with-parallel.md
```diff
@@ -1,39 +1,51 @@
...
-The simplest way is to use a double for-loop
-to iterate over the batches sequentially and collect them. pyDVL provides the simple convenience class
-[SequentialInfluenceCalculator][pydvl.influence.influence_calculator] to do this. The
-batch size should be chosen as large as possible, such that the corresponding batches fit
-into memory.
+    For speeding-up data valuation algorithms with parallelization
+    see [[speed-up-value-with-parallel]].
+
+
+### Sequential Computation
```
janosg, line 21: I think the sequential computation should be explained in a separate document that is marked as prerequisite for this one.

janosg (commented):
I think it's very good now. Just have two minor comments we can discuss later.

Issue #600, opened 2024-06-13
Simplify caching


janosg:
The new design of data valuation methods avoids repeated computations of the utility function without relying on caching. We could therefore get rid of our current caching implementation based on memcached, which seems overpowered. This would close several issues related to caching (e.g. #517, #475, #464 and #459). Moreover, it could solve problems that arise due to the many files the current caching solution creates. 

The only situation where caching ist still really important is when one benchmarks multiple algorithms and wants to use caching to ensure that randomness is kept as constant as possible between different algorithms and to save runtime in the benchmark. We therefore should create an entry point for benchmarking frameworks to enable caching. I see two possible solutions:

1. Use a simple shared-memory cache to store all utility evaluations and return them as part of the ValuationResult. A benchmarking library could then use these evaluations to build up a cache. All logic to wrap Utility with a cached version would be in the benchmarking library.
2. We could keep the `cache_backend` abstraction in the Utility but only implement a much simpler shared-memory backend in pydvl. Users with advanced caching needs could then build their own backends.

AnesBenmerzoug:
Now that we only use joblib for the parallelization of data valuation algorithms we could also leverage its caching mechanism through the [Memory](https://joblib.readthedocs.io/en/latest/memory.html) class and maybe only offer one extension to support caching in a distributed setting.

I tried using it when I refactored the caching backends and couldn't really make it work with memcached because it is implemented as a file-based caching. So I gave up on basing our code on it but I still took heavy inspiration from their interface so perhaps we could consider it again.

mdbenito:
I agree that without the reliance on caching for efficient evaluation of utilities over permutations, caching is not so useful for single methods. However, it's main benefit is then for experiments where one wishes to compare different methods, while isolating the randomness of the utility evaluations. Caching is perfect for that, and removing it would be a loss IMO. I think that what we need is to address #475 in a way that allows specifically for this (very) important use case.

PR #610, opened 2024-07-07
Refactor KNN Shapley

### LINKED ISSUE #613: Potential bug in KNN Shapley

janosg:
### Description

This PR refactors the KNN shapley valuations to be more aligned with the new Valuation classes. 

### Changes

- [x] Completely refactor KNN Shapley 
- [x] Add parallelization
- [x] Add more tests 

closes #613

### Checklist

- [x] Updated Changelog


src/pydvl/valuation/methods/knn_shapley.py
```diff
@@ -13,3 +13,143 @@
...
+    def fit(self, data: Dataset) -> Self:
+        """Calculate exact shapley values for a KNN model on a dataset.
+
+        This fit method bypasses direct evaluations of the utility function and
+        calculates the Shapley values directly.
+
+        In contrast to other data valuation models, the runtime increases linearly
+        with the size of the test dataset.
+
+        Calculating the least core valuation is a computationally expensive task that
```
schroedk, line 68: Looks like a copy-paste leftover (or is there any connection to least core?)

src/pydvl/valuation/methods/knn_shapley.py
```diff
@@ -13,3 +13,143 @@
...
+
+    Args:
+        utility: KNNUtility with a KNN model to extract parameters from. The object
+            will not be modified nor used other than to call [get_params()](
+            <https://scikit-learn.org/stable/modules/generated/sklearn.base.BaseEstimator.html#sklearn.base.BaseEstimator.get_params>)
+        progress: Whether to display a progress bar.
+
+    """
+
+    def __init__(self, utility: KNNClassifierUtility, progress: bool = True):
```
schroedk, line 49: Missing call to `super().__init__()`

src/pydvl/valuation/methods/knn_shapley.py
```diff
@@ -13,3 +13,143 @@
...
+        x: A single test data point.
+        y: The correct label of the test data point.
+        y_train: The training labels.
+
+    Returns:
+        The Shapley values for the test data point.
+
+    """
+    n_obs = len(y_train)
+    k = helper_model.get_params()["n_neighbors"]
```
schroedk, line 135: For better readability: `n_neighbors`?

src/pydvl/valuation/methods/knn_shapley.py
```diff
@@ -13,3 +13,143 @@
...
+    # sorts data indices from close to far
+    sorted_indices = helper_model.kneighbors(
+        x.reshape(1, -1), n_neighbors=n_obs, return_distance=False
+    )[0]
+
+    values = np.zeros(n_obs)
+
+    idx = sorted_indices[-1]
+    values[idx] = float(y_train[idx] == y) / n_obs
+    # reverse range because we want to go from far to close
```
schroedk, line 146: What about sorting `sorted_indices` in a descending way? This could simplify the for-loop (better to understand).

src/pydvl/valuation/methods/knn_shapley.py
```diff
@@ -13,3 +13,143 @@
...
+            status=Status.Converged,
+            values=values,
+            data_names=data.data_names,
+        )
+
+        self.result = res
+        return self
+
+
+def _compute_values_for_one_test_point(
```
schroedk, line 116: As in the previous PR, I would bundle this with the class in the form of a static method.

src/pydvl/valuation/methods/msr_banzhaf.py
```diff
@@ -172,6 +169,9 @@ def _combine_results(
         Returns:
             The combined valuation result.
 
+        TODO: Verify that the two running means are statistically independent (which is
+        assumed in the aggregation of variances).
```
schroedk, line 173:
```suggestion
 assumed in the aggregation of variances).
```

src/pydvl/valuation/scorers/knn.py
```diff
@@ -0,0 +1,40 @@
...
+            likelihoods = []
+            for i in range(len(y)):
+                if y[i] not in label_to_pos:
+                    likelihoods.append(0.0)
+                else:
+                    likelihoods.append(probs[i, label_to_pos[y[i]]])
+            return np.mean(likelihoods)
+
+        super().__init__(
+            scoring=scoring,
```
schroedk, line 35:
```suggestion
scoring=cast(SupervisedScorerCallable, scoring)
```

src/pydvl/valuation/scorers/knn.py
```diff
@@ -0,0 +1,40 @@
+"""Specialized scorer for k-nearest neighbors models."""
+
+
```
schroedk, line 3:
```suggestion
from typing import cast
```

src/pydvl/valuation/utility/knn.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def __init__(
+        self,
+        model: KNeighborsClassifier,
+        test_data: Dataset,
+        *,
+        catch_errors: bool = True,
+        show_warnings: bool = False,
+        cache_backend: CacheBackend | None = None,
+        cached_func_options: CachedFuncConfig | None = None,
```
schroedk, line 52: I am wondering, why this is seperated from CacheBackend. When I understand this correctly, the parameters of `CachedFuncConfig` can include specifics (like ignore args) of the function to wrap. In this case it is a `_utility`, which is not visible to the user. Maybe we can open an issue to think about, if it makes sense to design the cache in a different way. What do you think?

src/pydvl/valuation/methods/knn_shapley.py
janosg: Thanks!

src/pydvl/valuation/methods/knn_shapley.py
janosg: Thanks, added it.

src/pydvl/valuation/methods/knn_shapley.py
janosg: Was on the fence but changed it now.

src/pydvl/valuation/methods/knn_shapley.py
janosg: Tried that but the problem is that you would need a reverse enumerate too.

src/pydvl/valuation/methods/knn_shapley.py
janosg: Done, thanks!

src/pydvl/valuation/utility/knn.py
janosg: This is not specific to this PR. We could discuss it in the next meeting. You could also make a comment in #600

src/pydvl/valuation/methods/msr_banzhaf.py
```diff
@@ -172,6 +169,9 @@ def _combine_results(
         Returns:
             The combined valuation result.
 
+        TODO: Verify that the two running means are statistically independent (which is
```
janosg, line 172: Is this really the correct formatting for todos?

src/pydvl/valuation/methods/msr_banzhaf.py
```diff
@@ -172,6 +169,9 @@ def _combine_results(
...
+ assumed in the aggregation of variances).
```
janosg, line 173: Is this really the correct formatting for TODOs?

PR #618, opened 2024-09-02
Fix wrong value check in owen method

schroedk:
There was a strange check for the sampling probability in the Owen sampling of the old code section, which made a tests fail.


janosg (approved):
I'm surprised this had such a large effect and I don't know why the condition was in there.

PR #631, opened 2024-12-05
Fixes for grouped dataset

mdbenito:
### Description

This PR fixes inconsistencies in how grouped dataset operates with indices and allows methods to consistently access data either via "logical indices" (those kept by `Dataset`) or "data indices" (those corresponding to the data arrays themselves)

### Changes

- Introduced methods to work with logical and data indices
- Hid access to the raw data arrays behind method `data()` which takes logical indices as argument 
- Fixed the implementation of __getitem__ to return a (Grouped)Dataset as would be expected
- Added tests covering 100% of dataset.py

### Checklist

- [ ] Updated Changelog


mdbenito:
@AnesBenmerzoug I'm not sure you have much time for this, so I'll merge right away and let you review post-merge if you feel like it 😇

PR #665, opened 2025-04-05
Fix publish to testpypi

### LINKED ISSUE #635: Cannot upload to test.pypi
See this error: https://github.com/aai-institute/pyDVL/actions/runs/12468880329/job/34800937561#step:6:705

@janosg do you know who has access? The password I have for user appliedAI doesn't work any more (if it ever did, I'm not sure it's the same account as for the regular pypi?)

AnesBenmerzoug:
### Description

This PR closes #635 

It uses [trusted publishing](https://docs.pypi.org/trusted-publishers/) to publish packages to testpypi. This uses the [OpenID Connect (OIDC)](https://openid.net/connect/) standard to exchange short-lived identity tokens between a trusted third-party service (Github in this case) and PyPI (TestPyPI in our case) instead of relying on a secret long-lived token that we have to generate and keep in Github secrets.

We should also do the same for publishing to PyPI in a separate PR.

I kept the changes to a minimum and mostly copied the changes from one of my repositories (https://github.com/AnesBenmerzoug/langsfer/blob/main/.github/workflows/main.yml). I haven't tested the changes because if they work a new package would be published on testpypi from this branch, which is something we don't necessarily want.

We should consider switching from [bump2version](https://github.com/c4urself/bump2version) to [bump-my-version](https://github.com/callowayproject/bump-my-version) because it is no longer maintained and the newer package offers nice features such as getting the current version without any parsing or regular expressions.

### Changes

- Configured trusted publishing in testpypi for this repository.
- Created separate CI job for publishing package to testpypi.

### Checklist

- [ ] ~Updated Documentation (if necessary)~
- [ ] ~Updated Changelog~


mdbenito:
Nice. Also much better to have the docs and testpypi jobs separate. Once I merge a couple of bugfixes I will test this and release a dev package.

mdbenito (approved):

PR #668, opened 2025-04-08
Feature/667 zarr 3 support

### LINKED ISSUE #667: zarr < 3 and numcodecs
The latest numcodecs 0.16.0 [drops part of a deprecated api](https://github.com/zarr-developers/numcodecs/issues/721) and [breaks zarr < 3](https://github.com/zarr-developers/zarr-python/issues/2963). Causing [this](https://github.com/aai-institute/pyDVL/actions/runs/14328120730/job/40159746296)

Which makes me wonder: why do we pin zarr < 3? Is it possible to work around any breaking changes introduced with the major release without much effort?

schroedk:
### Description

This PR closes #667 

### Changes

- bump version dependency to zarr
- fix import due to version update

### Checklist

- [x] ~Updated Documentation (if necessary)~
- [x] Updated Changelog


mdbenito:
👍🏽 Although this is very mysterious. I have just had CI fail on the notebook tests because `requirements-notebooks.txt` pins zarr to 2.16 and there is no constraint on numcodecs, but the jobs on this PR haven't failed... 🤷🏽 

I think we should apply the same logic to `requirements-notebooks.txt` just in case.

mdbenito (commented):
👍🏽 Although this is very mysterious. I have just had CI fail on the notebook tests because requirements-notebooks.txt pins zarr to 2.16 and there is no constraint on numcodecs, but the jobs on this PR haven't failed... 🤷🏽

I think we should apply the same logic to requirements-notebooks.txt just in case.

mdbenito (approved):

PR #669, opened 2025-04-10
Switch to trusted publishing for pypi

mdbenito:
### Description

This PR mimics #665 for the main pypi repo

### Changes

- Ditches tokens and uses pypa's action
- Removes the twine dependency

### Checklist

- ~[ ] Wrote Unit tests (if necessary)~
- ~[ ] Updated Documentation (if necessary)~
- ~[ ] Updated Changelog~
- ~[ ] If notebooks were added/changed, added boilerplate cells are tagged with `"tags": ["hide"]` or `"tags": ["hide-input"]`~

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -23,8 +23,13 @@ env:
 jobs:
   publish:
     runs-on: ubuntu-22.04
+    environment:
+      name: publish
+      url: https://pypi.org/pydvl
     concurrency:
       group: publish
+    permissions:
+      id-token: write
     steps:
       - name: Checking out last commit in release
         if: ${{ github.event_name != 'workflow_dispatch' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yaml
```diff
@@ -54,6 +59,11 @@ jobs:
           # Make the version available as env variable for next steps
           echo CURRENT_VERSION=$CURRENT_VERSION >> $GITHUB_ENV
         shell: bash
+      - name: Build dist
+        run: |
+          python setup.py sdist bdist_wheel
+      - name: Publish to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
```
AnesBenmerzoug, line 66: Small suggestion: you could enable debug verbosity and printing of hashes, just in case.

https://github.com/pypa/gh-action-pypi-publish?tab=readme-ov-file#for-debugging

mdbenito: Good idea

PR #679, opened 2025-11-15
vibcoded partial fit

GrachevArtem:


GrachevArtem:
Hi 
Sorry we supposed to do this PR originaly in our private fork and then put it here. It is not ready yet and I forgot to delete. 
Are you interested in something like this?

